PR #7, opened 2016-03-13
fix #6: destructuring with a default parameter

### LINKED ISSUE #6: Error: Unsupported function signature.
Hey,

I think I'm running into a bug with this plugin. It fails with `Unsupported function signature.` everytime I use destructuring with a default parameter.

Error is:

```
/mocha-babel-flow/node_modules/eslint-plugin-flowtype/dist/utilities/getParameterName.js:30
    throw new Error('Unsupported function signature.');
```

Repro repo: https://github.com/vhf/mocha-babel-flow (clone && `npm run lint`)

It seems `function x({ y = 1 } = {}) { }` is enough to make the crash happen.

Am I missing something?

vhf:
Here's a quick fix for #6 .

### Tests
I added an "invalid" fixture but wasn't able to create a "valid" fixture.

My valid fixture was: `({foo = 1}: {foo: number} = {}) => {}'` which raises the following error:

```
    AssertionError: Should have no errors but had 1: [ { ruleId: null,
    fatal: true,
    severity: 2,
    source: '({foo = 1}: {foo: number} = {}) => {}',
    message: 'Parsing error: Binding rvalue',
    line: 1,
    column: 2 } ]
```

### Linting
There were a couple of errors and warnings when linting the plugin itself (`npm run lint`), but none of these were introduced by this patch so I decided to ignore them.

### DIFF
--- a/src/utilities/getParameterName.js
+++ b/src/utilities/getParameterName.js
@@ -16,9 +16,7 @@ export default (identifierNode, context) => {
     if (identifierNode.type === 'ObjectPattern' || identifierNode.type === 'ArrayPattern') {
         return context.getSourceCode().getText(identifierNode);
     }
-
-    if (_.has(identifierNode, 'left.type') && _.has(identifierNode, 'right.type') &&
-        identifierNode.left.type === 'ObjectPattern' && identifierNode.right.type === 'ObjectExpression') {
+    if (_.get(identifierNode, 'left.type') === 'ObjectPattern' && _.get(identifierNode, 'right.type') === 'ObjectExpression') {
         return context.getSourceCode().getText(identifierNode.left);
     }
 
--- a/src/utilities/getParameterName.js
+++ b/src/utilities/getParameterName.js
@@ -17,5 +17,10 @@ export default (identifierNode, context) => {
         return context.getSourceCode().getText(identifierNode);
     }
 
+    if (_.has(identifierNode, 'left.type') && _.has(identifierNode, 'right.type') &&
+        identifierNode.left.type === 'ObjectPattern' && identifierNode.right.type === 'ObjectExpression') {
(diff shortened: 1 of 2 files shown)


vhf:
Note: I expected Travis to fail here because of an incompatibility issue with `eslint@2.3.x`. In order to pass the tests locally I had to pin it to 2.2.0.

`npm install --save-dev --save-exact eslint@2.2.0` fixes the problem.

vhf:
@gajus : any feedback?

src/utilities/getParameterName.js
```diff
@@ -17,5 +17,10 @@ export default (identifierNode, context) => {
         return context.getSourceCode().getText(identifierNode);
     }
 
+    if (_.has(identifierNode, 'left.type') && _.has(identifierNode, 'right.type') &&
```
gajus:
```js
if (_.has(identifierNode, 'left.type') && _.has(identifierNode, 'right.type') && identifierNode.left.type === 'ObjectPattern' && identifierNode.right.type === 'ObjectExpression') {
```

can be simply written as:

```js
if (_.get(identifierNode, 'left.type') === 'ObjectPattern' && _.get(identifierNode, 'right.type') === 'ObjectExpression') {
```

gajus:
> Note: I expected Travis to fail here because of an incompatibility issue with eslint@2.3.x. In order to pass the tests locally I had to pin it to 2.2.0.

I cannot merge a PR if it is failing with the latest ESLint version. Latest version now is v2.4.0. Is it still failing?

vhf:
Thanks for the feedback, much appreciated! Travis is happy now. :)

There was a known issue with the latest eslint (babel-eslint folks were recommending pinning to eslint@2.2.0 until a fix was available), seems all good now!

PR #10, opened 2016-04-02
Use 'void' and 'undefined' interchangeably

joonhocho:
PR to allow 'void' in place of 'undefined'.


gajus:
Thank you. v2.2.3 has been published.

joonhocho:
thanks!

PR #12, opened 2016-04-05
issue 11: space-after-type-colon does not apply to the return type

johnsonw:
Given the following configuration:

```json
   1,
   "never"
]
```

and a function:

```js
function fn (key: string): Object {
```

Running the linter will produce the following:

> xxx:yy  warning  There must be no space after "key" parameter type annotation colon  flowtype/space-after-type-colon

This patch will address this issue and ensure that the return type is also taken into consideration when checking for spaces after the colon.

### DIFF
--- a/src/rules/spaceAfterTypeColon.js
+++ b/src/rules/spaceAfterTypeColon.js
@@ -24,5 +24,17 @@ export default iterateFunctionNodes((context) => {
                 }
             }
         });
+
+        if (functionNode.returnType) {
+            const spaceAfter = functionNode.returnType.typeAnnotation.start - functionNode.returnType.start - 1;
+
+            if (always && spaceAfter > 1) {
+                context.report(functionNode, 'There must be 1 space after return type colon.');
+            } else if (always && spaceAfter === 0) {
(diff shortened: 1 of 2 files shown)


src/rules/spaceAfterTypeColon.js
```diff
@@ -1,4 +1,5 @@
 import _ from 'lodash';
+import {format} from 'util';
```
gajus: Please remove `format` in favour of string concatenation (as it was).

Your are welcome to submit a separate PR promising to use `util.format`, though this addition is not required to solve this issue.

johnsonw:
Anything else I need to address in this patch for it to land? I can open a separate patch to handle space-before-type-colon.

gajus:
@johnsonw Please squash the commits. If you do not know how, here is a guide. http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

johnsonw:
Commits have been squashed.

gajus:
Thanks. Thats a lot clearer.

gajus:
v2.2.5 has been published.

PR #16, opened 2016-04-21
Properly handle arrow function expressions and generic return type annotations

### LINKED ISSUE #15: Shorthand arrow return appears to be except from linting rules
With this linting configuration:

```json
    "rules": {
        "flowtype/require-parameter-type": 2,
            2,
            "always",
            {
                "annotateUndefined": "never"
            }
        ],
            2,
            "always"
        ],
        "flowtype/space-before-type-colon": [
            2,
            "never"
        ],
    },
```

The following will not return a linting error:

```javascript
const getUserAgent = (state: Object) => state.device.userAgent;
```

The following will return a false positive:

```javascript
const func = (): Function => (
    () => {}
);
```

```
flowtype/require-return-type   Must not annotate undefined return type
```

The following also returns a false positive:

```javascript

type ISize = {
    width: number,
    height: number,
};
const getScreenDimensions = (state: Object): ISize => state.device.screenDimensions;
```

```
flowtype/require-return-type   Must not annotate undefined return type
```

justinanastos:
An arrow function can be written shorthand like so:

```
(foo) => "foo"
```

This was improperly being recognized as not returning a value. This will check to see if the arrow function is an expression and if it is, will be assumed to be returning something.

---

Before the node's return type was deemed undefined if it had one of the following:

- `typeAnnotation` of `VoidTypeAnnotation`
- `typeAnnotation` of `GenericTypeAnnotation`

That will produce false positives because a `typeAnnotation` of `GenericTypeAnnotation` can be things other than `undefined`. This will check for a `GenericTypeAnnotation` that is actually `undefined`.

Closes #15

### DIFF
--- a/src/rules/requireReturnType.js
+++ b/src/rules/requireReturnType.js
@@ -32,7 +32,8 @@ export default (context) => {
             throw new Error('Mismatch.');
         }
 
-        const isFunctionReturnUndefined = !targetNode.returnStatementNode || isUndefinedReturnType(targetNode.returnStatementNode);
+        const isArrowFunctionExpression = functionNode.expression;
+        const isFunctionReturnUndefined = !isArrowFunctionExpression && (!targetNode.returnStatementNode || isUndefinedReturnType(targetNode.returnStatementNode));
         const returnTypeTypeAnnotationType = _.get(targetNode, 'functionNode.returnType.typeAnnotation.type');
         const isReturnTypeAnnotationUndefined = returnTypeTypeAnnotationType === 'GenericTypeAnnotation' || returnTypeTypeAnnotationType === 'VoidTypeAnnotation';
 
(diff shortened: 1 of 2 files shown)


src/rules/requireReturnType.js
```diff
@@ -25,6 +25,39 @@ export default (context) => {
         return returnNode.argument === null || returnNode.argument.name === 'undefined' || returnNode.argument.operator === 'void';
     };
 
+    const getIsReturnTypeAnnotationUndefined = (targetNode) => {
+        const isReturnTypeAnnotationLiteralUndefined = _.isMatch(
```
gajus: I don't like the approach taken here (it is too verbose).

I would prefer this written as:

```js
_.get(targetNode, 'functionNode.returnType.typeAnnotation.id.name', 0) === undefined;
// ...
```

justinanastos: Sure thing.

justinanastos:
@gajus I removed the `_.match()` in favor of `_.get()`.

gajus:
A solid PR. Thank you, @justinanastos. `eslint-plugin-flowtype@2.2.7` has been published.

justinanastos:
Thank you @gajus ! This is an enormously helpful plugin, thank you for building it.

PR #26, opened 2016-06-26
require-valid-file-annotation rule implementation

BenoitZugmeyer:
A new rule to check for typos or placement error of the `// @flow` comment.

Even if I know the rule, I keep inadvertently inserting the `/*eslint ...*/` comment or `"use strict"` at the beginning of the file, disabling the flow check, and nothing is warning me about this. I think it would be a great addition to this plugin.

I'm open to discussion: you can consider this as a proposal and a prototype implementation.

PS: I love the `readmeAssertions.js` script! great idea.
PPS: You should commit the .eslintrc you are using for this project.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -86,6 +86,7 @@ When `true`, only checks files with a [`@flow` annotation](http://flowtype.org/d
 
 {"gitdown": "include", "file": "./rules/require-parameter-type.md"}
 {"gitdown": "include", "file": "./rules/require-return-type.md"}
+{"gitdown": "include", "file": "./rules/require-valid-file-annotation.md"}
 {"gitdown": "include", "file": "./rules/space-after-type-colon.md"}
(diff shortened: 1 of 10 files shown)


gajus:
> PPS: You should commit the .eslintrc you are using for this project.

It is inherited from https://github.com/gajus/pragmatist (which in turn gets it from https://github.com/gajus/canonical (which in turn gets it from https://github.com/gajus/eslint-config-canonical)). In theory, it should be enough to include https://github.com/gajus/eslint-config-canonical. But the idea behind Pragmatist was to reduce boilerplate code.

gajus:
v2.3.0 has been released

BenoitZugmeyer:
Great!

gajus:
Thank you for a great PR.

I have added ESLint configuration and fixed the CI.

PR #28, opened 2016-06-30
Fixes to README

ffxsam:
Corrected a mistake regarding type aliases

### DIFF
--- a/.README/rules/type-id-match.md
+++ b/.README/rules/type-id-match.md
@@ -17,6 +17,6 @@ This rule needs a text RegExp to operate with Its signature is as follows:
 }
 ```
 
-`'^([A-Z][a-z0-9]+)+Type$'` is the default pattern.
+`'^([A-Z][a-z0-9]+)+Type$$'` is the default pattern.
 
 <!-- assertions typeIdMatch -->
--- a/README.md
+++ b/README.md
@@ -437,24 +437,23 @@ This rule needs a text RegExp to operate with Its signature is as follows:
 }
 ```
 
-`'^([A-Z][a-z0-9]+)+Type
-` is the default pattern.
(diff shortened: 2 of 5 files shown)


ffxsam:
Sorry, I'm not really sure how to edit those files. I don't see any of the text that I changed.

gajus:
Sorry, provided wrong link. This is where the source files are, https://github.com/gajus/eslint-plugin-flowtype/blob/master/.README/README.md

ffxsam:
Are you sure [this](https://github.com/gajus/eslint-plugin-flowtype/blob/master/tests/rules/assertions/typeIdMatch.js) isn't the file I should edit?

Are you saying that editing that will modify the `README.md` file?

ffxsam:
There's nothing in `.README/README.md` that contains the code I had fixed originally.

gajus:
https://github.com/gajus/eslint-plugin-flowtype/tree/master/.README/rules

ffxsam:
Like I said, there's nothing in the rules folder that matches what I changed in my PR.

gajus:
It is https://github.com/gajus/eslint-plugin-flowtype/blob/master/tests/rules/assertions/typeIdMatch.js. You are right. 

After making changes you need to run `npm run readme` to build the docs.

Sorry, I am jumping between several projects.

ffxsam:
Ok, done! I also found a bug in `bin/readmeAssertions.js`. This section in particular was causing `npm run readme` to fail:

https://github.com/gajus/eslint-plugin-flowtype/blob/master/bin/readmeAssertions.js#L33-L36

`require(filePath).default` was `undefined`. Something must've changed with babel, maybe? The valid and invalid codes are exported as defaults, and so they can simply be accessed at the root level (`require(filePath).valid`).

gajus:
Thank you

PR #34, opened 2016-07-17
Allow require-valid-file-annotation to blacklist some file

NickMalt:
This is will be helpful in situation when we want to check all file without flow annotation, but in some case we should be able to blacklist some file.


gajus:
@NickMalt can you document a reason for withdrawing the proposal? Just to give context.

PR #46, opened 2016-08-09
Support for Optional Args in spaceBeforeTypeColon

danharper:
Just allowing an additional character (the `?`) when the type is optional.

```js
(foo?: string) => {}
```


gajus:
Perfect. Thank you! v2.4.1 published.

PR #47, opened 2016-08-09
Fix "space-before-type-colon" for non-arrow functions

### LINKED ISSUE #20: `space-before-type-colon` doesn't seems to work
`flowtype/space-after-type-colon` works well in my case, but it's before counterpart does not. 

I'll post a simple repro asap.

danharper:
Currently the `space-before-type-colon` rule only works for arrow functions. It seems the current implementation is actually based off a parsing bug: https://github.com/babel/babylon/issues/51 - for normal functions, the parameter locations include the type annotations, but they don't for arrow functions.

It now works for both normal functions and arrow functions (and should continue working if that bug is fixed). I've also added a few more tests to `space-after-type-colon` just to be sure it works.

Fixes #20.

### DIFF
--- a/src/rules/spaceBeforeTypeColon.js
+++ b/src/rules/spaceBeforeTypeColon.js
@@ -7,13 +7,15 @@ import {
 export default iterateFunctionNodes((context) => {
     const always = context.options[0] === 'always';
 
+    const sourceCode = context.getSourceCode();
+
     return (functionNode) => {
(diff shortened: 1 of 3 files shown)


gajus:
Thank you! v2.4.2 has been released.

danharper:
Cheers! 👍  I was just about to push the readme update when you merged, so that's here: #48

PR #49, opened 2016-08-09
Correct undefined for async & generator functions

### LINKED ISSUE #22: When using Async/Await annotating as Promise<void> does not work with annotateUndefined.
With the following:

```js
async function doThing(): Promise<void> {}
```

a promise is implicitly returned and flow requires that you annotate it as a promise.

If you have `{ "annotateUndefined": true }`, this plugin does not accept throws an error since `Promise<void>` is not `undefined`.

danharper:
Fixes #22.


gajus:
v2.4.3 has been released. Thank you!

Issue #41, opened 2016-08-01
REQ: Option to ignore anonymous functions for require-return-type rule


turadg:
I want `flowtype/require-return-type` enabled in my codebase for `function` declarations and concise methods but it also complains on,
```
      .then(profile => Profile.fromJSON(profile));
```
and complying is a little ugly,
```
      .then((profile): Profile => Profile.fromJSON(profile));
```

I figure that Flow can generally infer the type in such small function bodies as are common with `=>` and when used anonymously there's no value in hinting to the IDE.

gajus:
Fixed in https://github.com/gajus/eslint-plugin-flowtype/pull/50

gilbarbara:
Awesome!
But npm still shows 2.4.1 as the last version..

gajus:
@gilbarbara Thank you for drawing my attention to it. Forgot `npm publish` in the workflow.

gilbarbara:
👍 
I used to forget to publish to npm often.. Since you are using travis just add `on: tags:true` (and your the secure npm key) to the config.

Issue #23, opened 2016-06-07
Exempt arrow functions from requiring types


indigoviolet:
Arrow functions should be small enough that types aren't essential; it'd be nice to be able to exempt them from `flowtype/require-parameter-type` and `flowtype/require-return-type`

However, I would not want this to affect arrow functions that are set as class properties to make `bind` easier:

```
class Foo extends React.Component {
  instancemethod = (foo: Type) : ReturnType => {
     ....
  }
}

```

Is this a feasible request?

gajus:
`flowtype/require-parameter-type` could be updated to support a setting which controls which types of functions to apply the style.

allanhortle:
+1 on this. 

Making heavy use of chained methods with arrow functions gets really heavy handed when the same object is being passed through:

```js
return Maybe
    .fromNull(props.files)
    .filter((files: Array): Array => files.size)
    .map((files: Array): Array => {
        return files.map((file: Object): Number => file.size);
    });
```

gajus:
Fixed in https://github.com/gajus/eslint-plugin-flowtype/pull/52 thanks to @danharper. Released in v2.6.0.

Issue #42, opened 2016-08-02
Feature Request: Do not require type annotations for variables where the type can be asserted


sigurdsvela:
In a lot of cases the type of a variable can be asserted from the context

```
const myArrayOfString: String[] = ['foo', 'bar', 'bad'];
myArrayOfString.map((s) => {});
```

In the above example for instance, it can be asserted that `s` is of type 'string'.
My request is a rule to require type annotations only where the type can not be asserted from the context.

This might not be possible/to much effort, but I thought I'd drop the idea. Hopefully, information about what variables type can be asserted is something Flow makes easily accessible.

gajus:
Fixed in https://github.com/gajus/eslint-plugin-flowtype/pull/50

PR #50, opened 2016-08-09
Allow excluding arrow functions from require-param

danharper:
For #23.


danharper:
Also looks like it'd close #41 and #42.

gajus:
v2.5.0 has been released. Thank you. :-)

PR #51, opened 2016-08-09
Fix where async func would not require return type

danharper:
Realised I messed up with #49 - sorry!

It would not error on an async/generator function missing a return type.

Cleaned up the implementation a little so `isFunctionReturnUndefined` can't be true if it's a async/generator function (as they'd return at least a Promise or Generator)


gajus:
v2.5.1 has been released. Thank you. :-)

PR #53, opened 2016-08-10
Fix Destructuring with spaceBeforeTypeColon

danharper:
Noticed that destructuring in a function parameter wouldn't report correctly.


gajus:
v2.6.1 has been released. Thank you. :-)

PR #55, opened 2016-08-10
Object & Class Type Spacing

danharper:
Extends the existing `spaceAfterTypeColon` and `spaceBeforeTypeColon` rules to apply to type objects, and also to class properties like:

```js
type Foo = { name: string }

const x: { name: string } = { name: 'Dan' }

class Foo { name: ?string }
```

Probably best to view the [first three commits](https://github.com/gajus/eslint-plugin-flowtype/compare/8a96d6a...abf05b43f896106ade14dcadc9a7b71f7f1e8c21) and then the [final commit](https://github.com/gajus/eslint-plugin-flowtype/commit/23f19989bb0946dc332e12a134b96c0998db5ad8) separately. The final commit refactors the result of the previous commits to share code between the function param evaluator and class property evaluator. [Final source](https://github.com/gajus/eslint-plugin-flowtype/blob/23f19989bb0946dc332e12a134b96c0998db5ad8/src/rules/spaceAfterTypeColon.js).

### DIFF
--- a/README.md
+++ b/README.md
@@ -701,6 +701,20 @@ type X = { foo?:?string }
 
 type X = { foo?:  ?string }
 // Message: There must be 1 space after "foo" type annotation colon.
+
+class X { foo:string }
+// Message: There must be a space after "foo" class property type annotation colon.
+
+// Options: ["never"]
+class X { foo: string }
+// Message: There must be no space after "foo" class property type annotation colon.
+
+class X { foo:?string }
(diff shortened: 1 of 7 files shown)


gajus:
Thank you! v2.6.3 has been published.

danharper:
Cheers! Just to check, has it definitely been published? NPM's showing 2.6.1 still.

gajus:
It wasn't. It is now.

I will add `semantic-release` over the weekend to avoid issues such as this.

Issue #3, opened 2016-02-23
require-return-type must not require undefined type


gajus:

migueloller:
There's another issue where if you specify `excludeArrowFunctions: 'expressionsOnly'` it will not check function expressions.

For example:
```es6
const add = function add(x: number, y: number) {
  return x + y;
};
// doesn't report any errors
```

danharper:
Hm, I've added a test for this, and it does error for "Missing return type annotation.":

![image](https://cloud.githubusercontent.com/assets/510740/17605520/afb7c3f2-6012-11e6-8779-2016bdfe1a11.png)

migueloller:
I had forgotten to add `always` in the options array on my `.eslintrc`. Funny though, because `require-parameter-type` worked without it. Is `always` not on by default for `require-return-type`?

migueloller:
In the documentation it shows that `require-param-type` doesn't have an `always` option but `require-return-type` does. Given that both rules are so similar it's a bit confusing that one has it and the other one doesn't. It would make sense if none did IMO.

Also, I have a few more examples (currently failing) that should not error:
```es6
/* @flow */

/**
 *  'flowtype/require-parameter-type': ['error', 'always', {
 *    excludeArrowFunctions: 'expressionsOnly',
 *  }],
 *
 * error: Missing 'x' parameter type annotation.
 */
[].forEach((x) => {
  console.log(x);
});

/**
 *  'flowtype/require-parameter-type': ['error', {
 *    excludeArrowFunctions: 'expressionsOnly',
 *  }],
 *
 * error: Missing 'x' parameter type annotation.
 */
[].forEach((x) => {
  console.log(x);
});

/**
 *  'flowtype/require-parameter-type': ['error', 'always', {
 *    excludeArrowFunctions: true,
 *  }],
 *
 * error: Missing 'x' parameter type annotation.
 */
[].forEach((x) => {
  console.log(x);
});

/**
 *  'flowtype/require-parameter-type': ['error', {
 *    excludeArrowFunctions: true,
 *  }],
 *
 * This one seems to work!
 */
[].forEach((x) => {
  console.log(x);
});

/**
 *  'flowtype/require-parameter-type': ['error', 'always', {
 *    excludeArrowFunctions: 'expressionsOnly',
 *  }],
 *
 * error: Missing 'x' parameter type annotation.
 */
[].forEach((x) => x);

/**
 *  'flowtype/require-parameter-type': ['error', {
[...]

PR #59, opened 2016-08-11
Proposal: Rename expressionsOnly option to conciseOnly?

danharper:
See: https://github.com/gajus/eslint-plugin-flowtype/issues/3#issuecomment-239321316

The option's a little confusing as I meant it to mean "an arrow function with an expression as the body", but it's easily misunderstood as "a function expression" (although it is under the `excludeArrowFunctions` option).

I chose "concise" as that's what MDN refers to them as:

> [Arrow functions can have either a "concise body" or the usual "block body".](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Function_body)
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Function_body

### DIFF
--- a/.README/rules/require-parameter-type.md
+++ b/.README/rules/require-parameter-type.md
@@ -6,7 +6,7 @@ Requires that all function parameters have type annotations.
 
 You can skip all arrow functions by providing the `excludeArrowFunctions` option with `true`.
 
-Alternatively, you can want to exclude only function expressions (e.g. `x => x * 2`). Provide `excludeArrowFunctions` with `expressionsOnly` for this. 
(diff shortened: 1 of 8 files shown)


danharper:
(hit Cmd+Enter too soon)..

I've kept the previous option in place, and duplicated the tests to cover each.

I'm not sure of the best way right now to "deprecate" these things, so I've just flagged it in the test, and then I'm filtering them out of the README output.

danharper:
Perhaps this should be held for a while though, until we can review other potentional cases: https://github.com/gajus/eslint-plugin-flowtype/issues/3#issuecomment-239322067

danharper:
replaced with #76, don't feel there's much value yet in renaming the value.

PR #62, opened 2016-08-13
Autofix for space-before/space-after

danharper:
These rules are now auto-fixable with ESLint's `--fix` 😃


danharper:
NOTE: I've also changed the expected position of the `?` "optional" token in some cases to be consistent.

Before we skipped over the `?` token in normal functions, but not in function type annotations, nor object type annotations.

This was inconsistent, and it seems correct to actually count it as the "tokenBeforeType", and we can add another rule to enforce spacing around `?` tokens

gajus:
This is beautiful. Thank you. v2.7.0 has been released.

PR #64, opened 2016-08-16
Class properties fail in getParameterName

jlkalberer:
For some reason when I updated to the latest version, I started seeing the exception getting thrown in this file.  I have a class like this:

```
type State = {bar: string};
class Foo {
  state: State;
}
```
That was failing

### DIFF
--- a/src/utilities/getParameterName.js
+++ b/src/utilities/getParameterName.js
@@ -17,7 +17,7 @@ export default (identifierNode, context) => {
         return identifierNode.argument.name;
     }
 
-    if (identifierNode.type === 'ObjectTypeProperty' || identifierNode.type === 'FunctionTypeParam') {
+    if (identifierNode.type === 'ObjectTypeProperty' || identifierNode.type === 'FunctionTypeParam' || identifierNode.type === 'ClassProperty') {
         return context.getSourceCode().getFirstToken(identifierNode).value;
     }
 


danharper:
What version of babel-eslint are you running?

Currently a ClassProperty should be picked up by the [`key.name`](https://github.com/gajus/eslint-plugin-flowtype/blob/d872b171e1623b922231b5e7126654563eaff381/src/utilities/getParameterName.js#L13).

where an example looks like:

```
Node {
  type: 'ClassProperty',
  start: 12,
  end: 23,
  loc: 
   SourceLocation {
     start: Position { line: 1, column: 12 },
     end: Position { line: 1, column: 23 } },
  computed: false,
  key: 
   Node {
     type: 'Identifier',
     start: 12,
     end: 15,
     loc: SourceLocation { start: [Object], end: [Object], identifierName: 'bar' },
     name: 'bar',
     range: [ 12, 15 ],
     _babelType: 'Identifier' },
  static: false,
  typeAnnotation: 
   Node {
     type: 'TypeAnnotation',
     start: 15,
     end: 23,
     loc: SourceLocation { start: [Object], end: [Object] },
     typeAnnotation: 
      Node {
        type: 'StringTypeAnnotation',
        start: 17,
        end: 23,
        loc: [Object],
        range: [Object],
        _babelType: 'StringTypeAnnotation' },
     range: [ 15, 23 ],
     _babelType: 'TypeAnnotation' },
  value: null,
  range: [ 12, 23 ],
  _babelType: 'ClassProperty',
  parent: .......
```

danharper:
(btw the failing tests are unrelated to this)

jlkalberer:
I'm using 6.0.4 babel-eslint

I'm also using flow throughout my project so it could be that.  It's hard to pinpoint the exact cause because I'm not really familiar with how eslint works but it was a polymorphic class

```
class Foo<TModel> {
   _config: Config;

   constructor(config: Config) {
     this._config = config;
   }
}
```

I would try to get you a true repro but it's a very large project and it would be hard to replicate the issue.

danharper:
Was a specific rule failing, or was it an "Unsupported function signature" exception in the code? You shouldn't see the exception anymore as of v2.7.1, but you also won't see the property name in the error message.

If you could find the failing syntax that'd be great, because currently `type State = {bar: string}; class Foo { state: State; }` works fine.

jlkalberer:
Yeah, I was seeing "Unsupported function signature".  

I'm on 2.7.0 I believe so I'll try updating and see if that fixes things.

jlkalberer:
Yeah, this is fixed in the latest. Thanks!

PR #65, opened 2016-08-16
Handle unknown syntax

danharper:
We just encountered an error where the rule was throwing on non-Flow file because it didn't understand how to display a name for it (as it works that out before it evaluates the code)

two changes:

1. added support for destructuring where the right-side is a "MemberExpression": `class X { foo({ bar }: Props = this.props) {} }`
2. don't throw when we can't evaluate a name for the parameter, just provide "Unknown signature" as the name

### DIFF
--- a/src/utilities/getParameterName.js
+++ b/src/utilities/getParameterName.js
@@ -24,7 +24,7 @@ export default (identifierNode, context) => {
     if (identifierNode.type === 'ObjectPattern' || identifierNode.type === 'ArrayPattern') {
         return context.getSourceCode().getText(identifierNode);
     }
-    if (_.get(identifierNode, 'left.type') === 'ObjectPattern' && _.get(identifierNode, 'right.type') === 'ObjectExpression') {
+    if (_.get(identifierNode, 'left.type') === 'ObjectPattern') {
         return context.getSourceCode().getText(identifierNode.left);
     }
 
--- a/tests/rules/assertions/spaceBeforeTypeColon.js
+++ b/tests/rules/assertions/spaceBeforeTypeColon.js
@@ -459,6 +459,9 @@ export default {
             ]
         },
         {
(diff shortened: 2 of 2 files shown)


gajus:
> don't throw when we can't evaluate a name for the parameter, just provide "Unknown signature" as the name

What is the reasoning for this?

danharper:
We had something like this in our code:

```js
// @flow
type Props = {}
class X {
  props: Props
  foo = ({ bar }: Props = this.props) => {}
}
```

which is valid Flow code, and should pass the lint rules.

However `getParameterName` didn't know how to display a name for `{ bar }: Props = this.props`, so it was throwing.

It actually calls `getParameterName` before it decides to report an error, so it'd actually throw even if the syntax is valid or not.

Because it's throwing, you'd get an error like this:

image

So we _could_ just call `getParameterName` when we report an error. That'll work fine when the lint is OK, but if it fails, it'll throw that error message still. So to prevent the end-user from seeing the confusing stack trace, I'm just returning a placeholder for it.

(Idea being it shouldn't break builds just because it can't print a pretty name for something)

gajus:
> However getParameterName didn't know how to display a name for { bar }: Props = this.props, so it was throwing.

This sounds like a canonical example of when to throw an error. (Unexpected input.)

> So we could just call getParameterName when we report an error.

I'd prefer to go with this option.

> That'll work fine when the lint is OK, but if it fails, it'll throw that error message still.

My view of this is that `getParameterName` is supposed to return a (string) name of a parameter. Returning another string that does not meet this definition introduces an extra conditional case that needs to be catered for every time when using `getParameterName`. Using `throw` enables the API user to handle the error if he chooses to or fail the program as it should.

danharper:
Ok, I've removed the second commit. I'll open another issue/PR later to discuss it.

Issue #63, opened 2016-08-16
static class property does not parse as expected


jgrund:
Consider:

```
declare class Foo {
  static bar:number;
}
```

This will output:

> There must be no space after "static" type annotation colon

danharper:
Looks like it's on `declare class` only.

The AST of the `declare class`'s property is an `ObjectTypeAnnotation` as opposed to a `ClassProperty`, and the code isn't handling the `static` property.

Should be a simple fix, will look into it tomorrow evening.

danharper:
this was fixed in v2.7.1

PR #66, opened 2016-08-16
Handle declare class with static properties

danharper:
See #63.

Also added tests for classes for static properties, just to be sure they worked.


danharper:
looks like Babylon updated earlier and fixed the bug with arrow function type annotation locations, so the tests are currently failing on a fresh `npm install`

PR #69, opened 2016-08-16
"Unknown function signature" in messages

danharper:
Continuing from #65.

When a parameter name can't be determined (because the specific syntax hasn't yet been accounted for), currently it'll throw an error which bubbles up to this message to the user:

Instead, I propose it should just skip the name in the message.

(In the previous PR I was just returning a fallback string which wasn't too nice. Instead it's now returning `null`, and I've handled that case in each error message)

### DIFF
--- a/src/rules/requireParameterType.js
+++ b/src/rules/requireParameterType.js
@@ -2,7 +2,8 @@ import _ from 'lodash';
 import {
     getParameterName,
     isFlowFile,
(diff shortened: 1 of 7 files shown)


danharper:
(test failures are resolved with #68)

gajus:
Thank you. v2.7.1 has been published.

PR #74, opened 2016-08-19
Add support for @noflow to require-valid-file-annotation.

STRML:
See https://flowtype.org/blog/2015/12/01/Version-0.19.0.html#noflow

It's nice to not error on this so you explicitly exclude a file, but still require that every file is opted in or out.


gajus:
@danharper are you happy with the change?

danharper:
👍 lgtm

gajus:
v2.8.0 has been released.

PR #75, opened 2016-08-19
Fix parens in type aliases in space-after-colon

### LINKED ISSUE #71: space-after-type-colon doesn't recognize parens
This is incorrectly triggering the rule:

```js
type Foo = {
  barType: (string | () => void),
  // ^-- warning because of parens
};
```

danharper:
Fixes #71

The first two commits here just re-structure the "spacing" tests to keep valid/invalid rules for each supported node type together. So when adding support for a certain syntax (like the next commits do), it's grouped together.

Second two commits are the fixes.


gajus:
Thank you. :-)

gajus:
Version v2.7.2 has been released.

PR #76, opened 2016-08-19
Clarify excludeArrowFunction option in docs

danharper:
just the clarification part of #59, i don't feel there's much value yet in renaming the value.


gajus:
Amazing. :-)

gajus:
Version v2.7.2 has been released.

PR #77, opened 2016-08-20
New rule: spacing around generic type annotation parameters

danharper:
`generic-spacing` to enforce spacing around generic type annotation parameters, defaulting to "never" (no spacing)

e.g. `Promise<string>` vs `Promise< string >`


gajus:
Wunderbar! v2.11.0 has been released. :-)

PR #78, opened 2016-08-20
New rule: Space before generic type annotation bracket

danharper:
`space-before-generic-bracket` to enforce spacing before a generic type annotation parameter bracket, defaulting to "never" (no spacing)

e.g. `Promise <string>` vs `Promise<string>`

(first commit updates the readme from previous changes which missed it, second commit is the actual changes)


gajus:
v2.9.0 has been published. Thank you. 👍

PR #79, opened 2016-08-20
new rule: union & intersection spacing

danharper:
`union-intersection-spacing`

e.g. `string | number` vs `string|number`

not enforcing against multiple spaces so that it allows multi-line unions like:

```
type X =
| string
| number
```


gajus:
Amazing. v2.10.0 has been released.

Issue #27, opened 2016-06-26
Support for optional properties and parameters


cayuu:
I probably missed a setting somewhere, but do optional props/params fly with arrow funcs?

```js
(optionalProp ?: string) => { (optionalProp : string|void) }
//             ^ Unexpected token (null)
```

Writing the above as a normal ES5 function validates fine. The pure "maybe" form seems to validate fine as an ES6 arrow func: 

```js
(maybe : ?string) => { (maybe : string|void|null) }
```

https://flowtype.org/docs/builtins.html#null-and-void

emirotin:
Additionally: the `space-before-type-colon` rule doesn't seem to work with optional params.

danharper:
Optional args are now supported as of #46

gajus:
v2.4.2 has been released. Thanks to @danharper

emirotin:
I'm using v2.11.0 and I've started getting the errors again, like:

> 5:35  warning  There must be a space after "i" parameter type annotation colon  flowtype/space-after-type-colon

referring to this line:

> type TArrayPredicate<T> = (el: T, i?: number) => boolean

danharper:
Huh, thought I'd covered that case. There's a PR open for it now :)

gajus:
v2.11.1 has been released. Thank you.

emirotin:
Yeah, looks fine, thank you.

PR #82, opened 2016-08-22
Fix optional FunctionTypeParam space-after-colon

danharper:
see: https://github.com/gajus/eslint-plugin-flowtype/issues/27#issuecomment-241531712


gajus:
v2.11.1 has been released. Thank you.

Issue #92, opened 2016-08-29
"type-id-match" pattern forbidding abbreviations


leoselig:
The default pattern for the `type-id-match` rule currently features `^([A-Z][a-z0-9]+)+Type$` (also in the README).

This, for instance, does effectively not allow the use of acronyms.
E.g. the rule produced an error for me when I wanted to declare a type `BatchedCRMItemsType`

How about chaning the regular expression to: `^([A-Z][a-z0-9]*)+Type$`? (meaning to make the lower case sequences in the camel case pattern optional)

PR #93, opened 2016-08-29
Allow uppercase sequences in `type-id-match`

leoselig:
Changes the default regular expression used to verify the names
of declared types. Previously, patterns with uppercase sequences
(e.g. `MyABCType`) produced errors, hence disallow the use of
acronyms in types (e.g. `CRM`).

ref https://github.com/gajus/eslint-plugin-flowtype/issues/92


gajus:
Thank you for the PR. v2.12.1 has been released.

PR #95, opened 2016-08-31
adds new rule for semicolon linting after type aliases

cloudkite:
This adds a new rule for linting semi colon usage.

this only checks semicolons are used after type aliases ie

```
type Thing = {
  a: number; // ignored
  b: string; // ignored
}; // checks this
```

I think linting usage of semicolons to delimit type properties should probably be a seperate rules? I was thinking along the lines of:

`type-property-delimiter`  with option of `'comma'` or `'semi'`
`type-property-delimiter-dangle` this rule would not check if there is either a semi or comma on the last type property declaration. Could have the same options as http://eslint.org/docs/rules/comma-dangle


cloudkite:
named this semi to align with http://eslint.org/docs/rules/semi

however maybe `semi-after-type-alias` is better?

danharper:
👍

danharper:
(i think you'll need to prefix the commit message with `feat: ` for the changelog generator)

cloudkite:
@danharper cool, in case there is any feedback/additional changes might be easier todo this using github's squash and merge feature at time of merge?

gajus:
> @danharper cool, in case there is any feedback/additional changes might be easier todo this using github's squash and merge feature at time of merge?

I am leaning towards the use of merge commit. It keeps full commit log in the history.

What is your reason to prefer "squash and merge"?

As far as I understand, the trade-off is:

* squash and merge produces a commit log that can be re-used as a change log
* creating a merge request makes for a very long commit log; most of the entries in the change log would not make much sense without knowing the context

cloudkite:
@gajus have updated the commit with label `feat` 

FWIW my take is that by using squash and merge you as repository maintainers have more control over the commits that go into the master branch.

ie if my PR has 4 commit entries
- add new rule
- add tests for rule
- fix typo in docs for new rule
- make changes from feedback

you can squash and merge this into one commit
- feat: add rule for linting use of semicolons

github puts a link to the original PR in the squashed commit so you can go back to the PR and see the full history

gajus:
> FWIW my take is that by using squash and merge you as repository maintainers have more control over the commits that go into the master branch.

Thank you for sharing your thoughts.

The biggest drawback is that you lose the power of the tools such as https://atom.io/packages/git-time-machine.

It is extremely convenient to be able to see why/when a particular line(s) have been added/changed/removed.

In this case though these is one commit anyway. :-)

Thank you for the PR.

cloudkite:
Awesome thanks! 👍

gajus:
v2.12.0 has been released.

PR #96, opened 2016-09-01
Add new rule for warning against use of weak types

cloudkite:
Warns when using weak type annotations **any**, **Object** and **Function**

### DIFF
--- a/src/rules/noWeakTypes.js
+++ b/src/rules/noWeakTypes.js
@@ -1,4 +1,4 @@
-import {get} from 'lodash';
+import _ from 'lodash';
 
 const reportWeakType = (context, weakType) => {
   return (node) => {
@@ -14,7 +14,7 @@ const genericTypeEvaluator = (context) => {
   const weakTypes = ['Function', 'Object'];
 
   return (node) => {
-    const name = get(node, 'id.name');
+    const name = _.get(node, 'id.name');
     const isWeakType = weakTypes.indexOf(name) >= 0;
 
     if (isWeakType) {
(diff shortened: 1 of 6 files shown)


src/rules/noWeakTypes.js
```diff
@@ -0,0 +1,31 @@
+import {get} from 'lodash';
```
gajus: To keep inline with the rest of the codebase, please change this to:

```js
import _ from 'lodash';
```

We are using https://github.com/lodash/babel-plugin-lodash to rewire them in the build step.

gajus:
Thank you!

cloudkite:
👍 cheers!

danharper:
@gajus the latest version published to npm was 2.11.4

gajus:
@danharper Released version 2.14.0.

Looks like CI script is silently failing. I will investigate why.

danharper:
👍 Thanks

gajus:
@danharper Release script fixed and tested across multiple repositories.

Couto:
This is just a question so that newbies like me can have some clarification:

If `Function` is a weak type (no argue there) and should be avoided, what's the proper way of documenting something like a compose function (per example)? 

`const compose = (f:Function, g:Function):Function => (x:mixed): mixed => f(g(x));` 

I'm new to flow and types, so please bear with me. :)

cloudkite:
@Couto you'll probably get more responses over at https://github.com/facebook/flow.

I would approach this with a polymorphic function see https://flowtype.org/docs/functions.html#polymorphic-functions

```
type Func<T, R> = (t: T) => R 

function compose<In, Out, T>(f: Func<T, Out>, g: Func<In, T>): Func<T , Out>  {
   return x => f(g(x))
}
```

* you could do it without the `Func` type alias. added for clarity
* could be a arrow function. but find full functions are a bit easier to annotation especially when its a polymorphic functions

PR #99, opened 2016-09-05
feat: options to prevent specific types in `noWeakTypes`

danharper:
e.g. I'd like to initially prevent the "easier" types (`Object` and `Function`), and later start disabling `any`.

This allows me to configure which rules to enable:

```js
{
  any: false,
  Object: true,
  Function: true
}
```

to disable the `any` check, while keeping the `Object` and `Function` checks

_although I can't decide if that reads more as "disallow `any` types, allow `Object`/`Function` types" - the opposite_


gajus:
https://github.com/gajus/eslint-plugin-flowtype/tree/v2.15.0 released.

Issue #40, opened 2016-08-01
Rule request: enforce comma vs semicolon in Object properties


cjb:
Hi!  Would you accept a PR for a require-trailing-commas rule, enforcing this style?
```js
type Foo = {
  bar: string,
  baz: string,
}
```
Thanks!

gajus:
Absolutely.

cloudkite:
In addition to this would be nice to have a rule that enforces semi colon delimiters vs comma delimiters ie.

```
type Foo = {
  bar: string;
  baz: string;
}
```

vs

```
type Foo = {
  bar: string,
  baz: string,
}
```

@cjb Happy to pick this up if you are too busy? @gajus any suggestions for naming of these new rules?

cjb:
@cloudkite That'd be great, don't wait for me!

danharper:
I'd use eslint's comma-dangle rule as a starting point for the trailing commas.

cloudkite:
I was thinking two new rules along the lines of:

1) `type-property-delimiter` 
with option of `'comma'` or `'semi'`

2) `type-property-delimiter-dangle`
this rule would not check if there is either a semi or comma on the last type property declaration. Could have the same options as http://eslint.org/docs/rules/comma-dangle

cjb:
@cloudkite Those look good! But FWIW they don't add the feature in the issue description.

cloudkite:
@cjb

using a combination of both would be the equivalent of `require-trailing-commas`

`type-property-delimiter: [2, 'comma']`
`type-property-delimiter-dangle:  [2, 'always']`

would have the following result
```
type Foo = {
  baa: number, // is fine
  bar: string; // error here as delimiter is semi instead of comma
  baz: string  // error here as no dangling/trailing comma
}
```

the reason for separating them would be to support people who prefer semi colons ie

`type-property-delimiter: ['semi']`
`type-property-delimiter-dangle:  ['always']`

```
type Foo = {
  baa: number; // is fine
  bar: string, // error here as delimiter is comma instead of semi
  baz: string  // error here as no dangling/trailing semi
}
```

cjb:
@cloudkite Cool, thanks, I see now!

danharper:
just found `eslint-plugin-babel` has a `babel/flow-object-type` rule for semicolon vs comma delimiters: https://github.com/babel/eslint-plugin-babel/blob/master/rules/flow-object-type.js

cloudkite:
[...]

PR #101, opened 2016-09-07
feat: add `delimiter-dangle` for trailing commas/semicolons

danharper:
See https://github.com/gajus/eslint-plugin-flowtype/issues/40

Trailing delimiters in `ObjectTypeAnnotation` and `TupleTypeAnnotation`.

To mirror ESLint's `comma-dangle` rule, it has the following options: `never` (default), `always`, `always-multiline`, `only-multiline`.

For the multiline rules, I'm pretty sure I'm correctly mirroring the logic of `comma-dangle` - it's multiline if the last property & closing brace are on separate lines.

For `ObjectTypeAnnotation`, I haven't added in the option of comma vs semi-colon. Another rule can handle that (hopefully the one from `eslint-plugin-babel`)

### DIFF
--- a/src/rules/delimiterDangle.js
+++ b/src/rules/delimiterDangle.js
@@ -1,0 +1,76 @@
+import _ from 'lodash';
+
+export default (context) => {
+  const option = context.options[0] || 'never';
+  const sourceCode = context.getSourceCode();
+
+  const reporter = (node, message, fix) => {
+    return () => {
+      context.report({
+        fix,
+        message,
+        node
+      });
+    };
+  };
+
+  const makeReporters = (node, tokenToFix) => {
+    return {
+      dangle: reporter(node, 'Unexpected trailing delimiter', (fixer) => {
+        return fixer.replaceText(tokenToFix, '');
+      }),
+      noDangle: reporter(node, 'Missing trailing delimiter', (fixer) => {
+        return fixer.insertTextAfter(tokenToFix, ',');
+      })
+    };
+  };
+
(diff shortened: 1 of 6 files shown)


danharper:
NOTE: don't merge yet. Just ran this on a real project (usually do this before a PR), and found a couple possible edge cases.

danharper:
🎉 fixed. @gajus thoughts?

src/rules/delimiterDangle.js
```diff
@@ -0,0 +1,95 @@
...
+
+    const [penultimateToken, lastToken] = sourceCode.getLastTokens(node, 2);
+
+    const isDangling = [';', ','].indexOf(penultimateToken.value) > -1;
+    const isMultiLine = penultimateToken.loc.start.line !== lastToken.loc.start.line;
+
+    const report = makeReporters(lastChildNode, penultimateToken);
+
+    if (option === 'always' && !isDangling) {
+      report.noDangle();
```
gajus: To avoid unnecessary if condition checks, each successful if condition should return, i.e.,

```js
if (option === 'always' && !isDangling) {
	report.noDangle();

	return;
}
```

gajus:
@danharper small comment (https://github.com/gajus/eslint-plugin-flowtype/pull/101#discussion_r77898497). Otherwise, LP(erfect)TM.

danharper:
Cool, just fixed that - I've rebased it into the initial commit.

gajus:
Just in case, go ahead and merge all the changes that you are comfortable with. :-)

PR #103, opened 2016-09-09
Add recommended rules preset (fixes #102)

### LINKED ISSUE #102: [suggestion] add "recommended" configuration
It'd be nice if this package contained "recommended" configuration that a users custom config could extend from, similar to `eslint:recommended` or [plugin:react/recommended](https://www.npmjs.com/package/eslint-plugin-react) or [plugin:lodash/recommended](https://www.npmjs.com/package/eslint-plugin-lodash).
The readme entry for each rule could then say whether it's included in the recommended config or not.

gajus:


danharper:
Few thoughts:

Remove `require-valid-file-annotation` and `no-weak-types` - you're likely adding Flow to an existing project, and at most it should just warn.

Remove `require-parameter-type` and `require-return-type` - same reasoning as above, and it kinda goes against Flow's type inference.

Remove `type-id-match` - overly specific.

Remove `delimiter-dangle` and `semi` - ESLint's recommended preset doesn't include it.

That just leaves us with enforcing common styling rules regarding spacing around symbols (e.g. I've never Flow code in the wild where you use `x :bool|2` vs `x: boolean | 2`), which seems sensible.

gajus:
@danharper Please merge this if you are happy with the changes.

danharper:
👍

gajus:
Thank you

PR #109, opened 2016-09-12
feat: add rule `no-dupe-keys` (fixes #108)

### LINKED ISSUE #108: Add rule to disallow duplicate type keys
Similar to http://eslint.org/docs/rules/no-dupe-keys

Bad
```javascript
type PropsType = {  
  data: Object,  
  currentValue: string,  
  data: Object,  
  onChange: Function,  
};
```
Good
```javascript
type PropsType = {
  data: Object,
  currentValue: string,
  onChange: Function,
};
```

aaron-harvey:


gajus:
@aaron-harvey please rebase onto the master branch and re-submit the PR.

src/rules/noDupe.js
```diff
@@ -0,0 +1,32 @@
+import forEach from 'lodash/forEach';
```
gajus: Please import the entire lodash bundle.

We are using `babel-plugin-lodash` to rewire dependencies in a build step.

danharper:
🎉  thanks for this :)

However currently it'll fail to report `type FooType = { a: number, a: string }`, as it's comparing the property source `a: number` with `a: string`. Could you check just the key instead?

For that it looks like you'd run into this `babel-eslint` bug too: https://github.com/babel/babel-eslint/issues/307 as there's currently no way to access the key of the property directly from the AST 😞 - you could probably parse out the raw tokens instead e.g. with `sourceCode.getFirstToken(node)` etc.

aaron-harvey:
I've applied the lodash change, changed from checking the line to the property, and added a test to cover `type FooType = { a: number, a: string }` case. :tada: 

Also, sorry for any alert spam.. I messed up the rebase :disappointed:

danharper:
This looks good, thanks :) 👍 Would you mind renaming it to `no-dupe-keys` to match ESLint's rule?

gajus:
Thank you @aaron-harvey

danharper:
Released as 2.18.0 - thanks! :)

gajus:
This continuous release is literally saving me hours every week. :-)

danharper:
Ha yeah, it's pretty awesome :)

PR #111, opened 2016-09-12
fix: copy recommended configs on build step

unRob:
Avoids throwing `Module build failed: Error: Failed to load plugin flowtype: Cannot find module './configs/recommended.json'`. Great job on this plugin!


danharper:
Thanks for this! 👍 (way better than what I was about to write). I've merged it in #112

PR #132, opened 2016-10-17
simple typo in readme (@floww)

akupila:


danharper:
Thanks for the PR, but the typo's intended - it's showing an example of the sort of typo the rule catches 🙂

akupila:
Oh hah, somehow i missed that :)

PR #140, opened 2016-10-23
Spacing cleanup

danharper:
`spaceBeforeTypeColon` and `spaceAfterTypeColon` contained a lot of duplication in reporting, but they also shared a lot of code with each other.

This splits all the shared logic, and the rules themselves are just a light wrapper around it. The gist is, for each node type we just call report with the colon token. The reporter then just finds the spacing between either the previous or next node.

Couple more cleanups I want to do, then will merge.


danharper:
Inadvertently, this also adds checking of some return types which we weren't checking before.

PR #18, opened 2016-04-21
Allow require-parameter-type to take ignoreParameterNames option

justinanastos:
Add an optional configuration parameter be added for `require-parameter-type` to allow a whitelist of parameter names that will be ignored for typechecks.

```json
{
    "parser": "babel-eslint",
    "plugins": [
        "flowtype"
    ],
    "rules": {
        "flowtype/require-parameter-type": [
            1,
            "always",
            {
                "ignoreParameterNames": [
                    "_"
                ]
            }
        ],
        "flowtype/require-return-type": [
            1,
            "always",
            {
                "annotateUndefined": "never"
            }
        ],
        "flowtype/space-after-type-colon": [
            1,
            "always"
        ],
        "flowtype/space-before-type-colon": [
            1,
            "never"
        ],
        "flowtype/type-id-match": [
            1,
            "^([A-Z][a-z0-9]+)+Type$"
        ]
    },
    "settings": {
        "flowtype": {
            "onlyFilesWithFlowAnnotation": false
        }
    }
}
```

#17

PR #143, opened 2016-10-25
feat: Add excludeParameterMatch option to require-parameter-type

### LINKED ISSUE #17: Feature Request: Allow whitelist of parameter names to be ignored by require-parameter-type
I propose an optional configuration parameter be added for `require-paremeter-type` to allow a whitelist of parameter names that will be ignored for typechecks.

```json
{
    "parser": "babel-eslint",
    "plugins": [
        "flowtype"
    ],
    "rules": {
        "flowtype/require-parameter-type": [
            1,
            "always",
            {
                "ignoreParameterNames": [
                    "_"
                ]
            }
        ],
            1,
            "always",
            {
                "annotateUndefined": "never"
            }
        ],
            1,
            "always"
        ],
        "flowtype/space-before-type-colon": [
            1,
            "never"
        ],
        "flowtype/type-id-match": [
            1,
            "^([A-Z][a-z0-9]+)+Type$"
        ]
    },
    "settings": {
        "flowtype": {
            "onlyFilesWithFlowAnnotation": false
        }
    }
}
```

I am using [`reselect`](https://www.npmjs.com/package/reselect) in a project. Selector functions take 2 parameters, a `state` and `ownProps`. For some selectors, I don't need the first parameters so I call the function like so, ignoring the `_` variable:

```
const getVideoRoot = (_, props: Object): string => props.videoRoot;
```

I would like to avoid typechecks on the `_` parameter because I don't use it.

I would be happy to submit a PR for this if it is something you'd be willing to merge in.

ahstro:
This adds the ability to include an option - `excludeParameterMatch` - that takes a regex of parameters to ignore.

This solves the same need as #17/#18, but is (in my opinion) a bit more flexible. (Also, there was no activity on that PR for a few months and I want this ;) )

Closes #17 
Closes #18

### DIFF
--- a/.README/rules/require-parameter-type.md
+++ b/.README/rules/require-parameter-type.md
@@ -32,4 +32,22 @@ Alternatively, you can want to exclude only concise arrow functions (e.g. `x => 
 }
 ```
 
+You can exclude parameters that match a certain regex by using `excludeParameterMatch`.
+
+```js
+{
+    "rules": {
+        "flowtype/require-parameter-type": [
+            2,
+            {
+              "excludeParameterMatch": "^_"
+            }
+        ]
+    }
+}
+```
+
+This excludes all parameters that start with an underscore (`_`).
+The default pattern is `a^`, which doesn't match anything, i.e., all parameters are checked.
+
 <!-- assertions requireParameterType -->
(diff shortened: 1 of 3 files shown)


danharper:
Awesome, thanks 👍

PR #144, opened 2016-10-25
Support dynamic accessors and callable objects in object-type-delimiter

zertosh:
I could not figure out how to rebuild the documentation, and had to `git commit --no-verify`. I would love to contribute more to this project, but the setup is so complicated and it's not very debug friendly :(

### DIFF
--- a/src/rules/objectTypeDelimiter.js
+++ b/src/rules/objectTypeDelimiter.js
@@ -41,6 +41,8 @@ const create = (context) => {
   };
 
   return {
+    ObjectTypeCallProperty: requireProperPunctuation,
+    ObjectTypeIndexer: requireProperPunctuation,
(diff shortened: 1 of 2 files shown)


danharper:
👍 Thanks, I'll update the contributing docs. I agree it's annoying - gets me every time too.

Basically, commit messages have to be prefixed with `fix: ` or `feat: ` or `refactor: ` etc. which determines the version bump automatically.

Makes deployment easier, but dev isn't so nice..

danharper:
Oh, and the docs are built automatically by Travis.

zertosh:
Oh, the problem was with the commit message - so strict :p Just force-pushed a corrected one

danharper:
Thanks, it's deploying now as 2.23.1 :)

Issue #128, opened 2016-10-12
Not all rules respect onlyFilesWithFlowAnnotation


BernsteinA:
In fact, it appears that only requireParameterType, requireValidFileAnnotation, and requireReturnType do.
Of those, requireValidFileAnnotation probably shouldn't anyway.

dikmax:
Looks like `requireValidFileAnnotation` check is wrong.
Right now it shows errors in files with `@noflow` annotation. When this annotation is removed, errors disappears too.

danharper:
Should be sorted now thanks to #155.

PR #155, opened 2016-11-18
refactor: improve flow file check logic

aaron-harvey:
This PR introduces a wrapper around each rule in the exported rules object that handles the `isFlowFile` and  `onlyFilesWithFlowAnnotation` logic.

This should resolve a handful of issues where some rules were not honoring the `onlyFilesWithFlowAnnotation` setting (e.g. sort-keys, type-id-match, #128).

I'm not sure if this is the best way to approach this problem, but it seemed simpler than having the checks in every rule file.

Also, there currently is no way for a rule to 'opt out' of this wrapper - I'm not sure if this is needed but it wouldn't be difficult to add that functionality if that need arises.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+import {checkFlowFileAnnotation} from './utilities';
 import defineFlowType from './rules/defineFlowType';
 import genericSpacing from './rules/genericSpacing';
 import noWeakTypes from './rules/noWeakTypes';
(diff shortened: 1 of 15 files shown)


danharper:
This is awesome, thanks!

I've fixed the conflicts & merged separately (https://github.com/gajus/eslint-plugin-flowtype/commit/52b71f3cccd4e0208e3d180d34d3769a92cc7d7f, https://github.com/gajus/eslint-plugin-flowtype/commit/d311f229132ae54ec3a5c293000bedab7966a5a1) - couldn't work out how to push to a PR.

Issue #152, opened 2016-11-12
Remove valid-syntax rule or add the meta deprecated flag


wagerfield:
Please could you either remove the `valid-syntax` rule or add `context.meta = { deprecated: true }`

Currently the only place that this rule is documented as being deprecated is in the [`README`](https://github.com/gajus/eslint-plugin-flowtype#valid-syntax)—it would be great to officially recognise this rule as deprecated using the context `meta` object.

gajus:
@wagerfield Please raise a PR.

danharper:
`deprecated: true` added with #156.

PR #156, opened 2016-11-19
chore: deprecate valid syntax rule

aaron-harvey:
This PR adds the `deprecated` property to the exported rule object (#152).


danharper:
Thanks.

PR #158, opened 2016-11-21
feat: add no-primitive-constructor-types rule

aaron-harvey:
This PR adds a rule that discourages the use of `Number`, `Boolean`, and `String` constructor types in favor of their primitive type counterparts.

Good: `type x = number`
Bad: `type x = Number`

From flowtype docs:

> Note that number and Number are separate types. The former is the type of primitive numbers which appear in programs as literals, like 3.14 and 42, or as the result of expressions like parseFloat(input.value). The latter is the type of Number wrapper objects, which are rarely used.

[More info on primitive types](https://flowtype.org/docs/builtins.html)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -15,6 +15,7 @@ import validSyntax from './rules/validSyntax';
 import booleanStyle from './rules/booleanStyle';
 import delimiterDangle from './rules/delimiterDangle';
 import noDupeKeys from './rules/noDupeKeys';
+import noPrimitiveConstructorTypes from './rules/noPrimitiveConstructorTypes';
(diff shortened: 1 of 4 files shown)


danharper:
Thanks :) The regex was disallowing custom types like `MyString`, so I've fixed that (https://github.com/gajus/eslint-plugin-flowtype/commit/ceea11c48f65f9d16032e4a673ae010736a56aec) and added docs (https://github.com/gajus/eslint-plugin-flowtype/commit/0d2e10f2c2ad84eaed0e0f913929483daa7df638). Deploying now.

aaron-harvey:
Good catch!

PR #170, opened 2016-12-07
feat: `require-return-type` include / exclude filter support (fixes #149)

### LINKED ISSUE #149: Whitelist / blacklist function names for require-return-type et al.
Feature request: being able to constrain `require-return-type` to only a set number of functions, based on their name.

In this particular case, I would like to enforce explicit return types for the `mapStateToProps` function used in react-redux. Knowing that this function is consistently named throughout my project, whitelisting the name in my `require-return-type` config would do the trick. Something along these lines:

``` json
  1,
  "always",
  {
    "includeOnlyMatching": [
      "mapStateToProps"
    ]
  }
],
```

Similarly, I would assume that there exist many use cases for the reverse; enabling the rule for all but a specific list of ignored function names. Something like `excludeMatching`?

Also, this kind of whitelist / blacklist functionality would likely make sense for some other rules as well, such as `require-parameter-type`.

Some related issues are #124 and #57. I don't think either would be solved by this feature, but it could serve as a workaround for situations where all-or-nothing won't work.

aaron-harvey:
This PR addresses #149 - adding the ability to include and exclude functions by name or regex patterns to the `require-return-type` rule.

Ex:
```js
{
    "rules": {
            2,
            "always",
            {
              "includeOnlyMatching": [
                  "Ba(r|z)"
              ]
            }
        ]
    }
}

{
    "rules": {
            2,
            "always",
            {
              "excludeMatching": [
                  "^F.*",
              ]
            }
        ]
    }
}

```

### DIFF
--- a/.README/rules/require-return-type.md
+++ b/.README/rules/require-return-type.md
@@ -34,4 +34,41 @@ Alternatively, you can want to exclude only concise arrow function (e.g. `() => 
 }
 ```
 
+You can exclude or include specific tests with the `includeOnlyMatching` and `excludeMatching` rules.
+
+```js
+{
+    "rules": {
+        "flowtype/require-return-type": [
+            2,
+            "always",
+            {
+              "includeOnlyMatching": [
+                  "^F.*",
+                  "Ba(r|z)"
+              ]
+            }
+        ]
+    }
+}
+
+{
+    "rules": {
+        "flowtype/require-return-type": [
+            2,
+            "always",
+            {
+              "excludeMatching": [
+                  "^F.*",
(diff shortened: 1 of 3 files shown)


patrikholcak:
Would love to see this get merged soon. I’m currently using my [old, sloppy implementation](https://github.com/patrikholcak/eslint-plugin-flowtype/commit/3c69fe5301fcbac98a2992fa70ecf82151ee11bb) which seems to work but I’m not happy about it being so outdated.

danharper:
Sorry, I missed this PR. Will merge later today! :)

PR #175, opened 2017-01-01
feat: add flowtype-errors

amilajack:
@gajus can you test if this works by linking? I'm getting an error saying that `Definition for rule flowtype-errors/show-errors was not found` if i enable `flowtype-errors/show-errors was not found` in my `eslintrc` when using the linked version of this plugin. I haven't worked with a `recommended.json` before I'm not sure how to debug this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,9 +19,11 @@
   },
   "license": "BSD-3-Clause",
   "peerDependencies": {
-    "eslint": ">=2.0.0"
+    "eslint": ">=2.0.0",
+    "flow-bin": "^0.37.4"
   },
   "dependencies": {
+    "eslint-plugin-flowtype-errors": "^2.0.3",
     "lodash": "^4.15.0"
   },
   "scripts": {
--- a/src/configs/recommended.json
+++ b/src/configs/recommended.json
@@ -34,8 +34,12 @@
       "always"
     ],
     "flowtype/use-flow-type": 1,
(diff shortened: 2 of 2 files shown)


amilajack:
Accidentally deleted the branch for this 😂

danharper:
I'm unsure about enabling it by default for `recommended`, as I'm sure many people are running `flow` separately anyway. Thoughts @gajus?

amilajack:
> I'm unsure about enabling it by default for recommended,

It is actually disabled in `recommended` config.

gajus:
I am not convinced that adding `eslint-plugin-flowtype-errors` just for the sake of being able to enable it via `flowtype-errors/show-errors` configuration is a valid use case.

To be clear, I am not suggesting to have it enabled by default either. As @danharper said, I too think that most people run separately.

@amilajack I don't see the value of making users to download `eslint-plugin-flowtype-errors` as a dependency of `eslint-plugin-flowtype`. Therefore, I am leaning towards simply documenting the existence of `eslint-plugin-flowtype-errors` in the `README.md`.

If you are Okay with this, please update to restrict changes to the `README.md` file. I am suggesting adding a section "Related projects" after the "Rules" section and documenting your plugin.

amilajack:
Sure thing. I'll close this PR and create a new one for the readme update

gajus:
@danharper I hope thats Okay with you too.

PR #180, opened 2017-01-11
feat: disallow flow annotations in files missing flow declaration (#176)

### LINKED ISSUE #176: Disallow flowtypes without @flow annotation
We're considering migrating a large codebase to use flow over time. As we migrate files, we'll add the `@flow` comment to files that we annotate. Today, it's easy to forget to add the `@flow` comment. When this happens, the type annotations appear valid, even though they aren't enforced. It would be nice to disallow flow annotations if the `@flow` comment is not defined.

aaron-harvey:
This PR adds a rule that disallows Flow annotations, imports, and alias in files that are missing a valid `@flow` declaration.

Closes #176


tests/rules/assertions/noTypesMissingFileAnnotation.js
```diff
@@ -0,0 +1,29 @@
...
+      }]
+    },
+    {
+      code: 'function t<T>(): T{}',
+      errors: [{
+        message: 'Type annotations require valid Flow declaration.'
+      }]
+    }
+  ],
+  valid: []
```
danharper: could you add some valid examples? e.g. ones containing `// @flow` and `// @noflow`?

aaron-harvey: I added a valid example.  I'm not 100% sure how to handle `// noflow` in this scenario - as of now the following is considered invalid:
```javascript
/* @noflow */
type FooType = number;
```

Thoughts?

danharper:
I think files with `@noflow` should allow type annotations as it's an explicit decision to temporarily opt-out of type checking in the file. Whereas a missing pragma, with type annotations, is usually likely a mistake.

micaste:
Hi @aaron-harvey, are you planning on finishing the PR ? I am ready to take over if needed.

For not checking `@noflow` files, you could create a util `stopOnNoFlowFileAnnotation.js` (to match the existing `checkFlowFileAnnotation.js`), and do something like:
```
if (key === 'no-types-missing-file-annotation') {
 return _.partial(stopOnNoFlowFileAnnotation, rule);
}
```

tibdex:
Hi @micaste and @aaron-harvey, this is a nice feature preventing silly mistaked. I am eager to use it! Any chance for one of you to tackle the changes requested by Dan ?

gajus:
Closing in favour of https://github.com/gajus/eslint-plugin-flowtype/pull/222.

PR #182, opened 2017-01-12
feat: support arrow function type shorthand

aaron-harvey:
This PR adds support for arrow expression type shorthand for params and return type.  I believe this is a valid way to declare arrow types (flow isn't complaining about it, and flow-ide picks it up).

It also includes a few minor performance enhancements, ie, moving some context things outside of the forEach loop.

```javascript
// example of what is now supported
type FnType = (a: string, b: number) => number
const arrowFunction: FnType = (a, b) => { return 42; }
```


aaron-harvey:
@danharper I think this is a good idea - I hadn't considered the case where a type was defined elsewhere. 
It also simplifies the code.

aaron-harvey:
Good catch.  I simplified the check and added a test to cover types in another file.

gajus:
> Good catch. I simplified the check and added a test to cover types in another file.

Is this still applicable?

PR #183, opened 2017-01-13
feat: add 'strict' eslint config

amilajack:


amilajack:
Accidentally deleted the branch for this 😂

amilajack:
>  I just wouldn't really want people to think just because it's the strict config, it's something to aim for.

That's what the recommended config is for. Strict is supposed to intent that everything is an error. No warnings. Also rules in the strict config can be overridden by users in their rules section of the eslint config.

gajus:
> require-parameter-type and require-return-type can be pretty verbose.

Agreed. I have tried enforcing these rules on a large code base and I don't recommend anyone to follow the suit.

> type-id-match

I wouldn't like this rule added either. It is too biased. Something like `type-id-match` would belong in one the style configuration libraries such as https://github.com/feross/standard or https://github.com/gajus/eslint-config-canonical (In fact, [I do enforce it in my own config](https://github.com/gajus/eslint-config-canonical/blob/f29f7c86a8e1c95c51c0655c27889e0b2fd62935/configurations/flowtype.json#L45-L48)).

amilajack:
So `type-id-match: 0, require-parameter-type: 0, require-return-type: 0`?

gajus:
> So type-id-match: 0, require-parameter-type: 0, require-return-type: 0?

These are off by default. Therefore, no need to include them.

amilajack:
Made the discussed changes

amilajack:
Safe to merge?

gajus:
> Safe to merge?

It is up to @danharper.

In general, is there a reason you don't want to publish this as a separate config?

Problem is, "strict" is an arbitrary term, so is the rule selection.

PR #194, opened 2017-02-06
Fix typo

glennreyes:


gajus:
@glennreyes README.md is a generated file. Please update https://github.com/gajus/eslint-plugin-flowtype/tree/master/.README instead.

glennreyes:
Oh thanks! Sure will do 👍

PR #197, opened 2017-02-20
fix: Ensure colon is not null in typeColonSpacing reporter

### LINKED ISSUE #196: "TypeError: Cannot read property 'value' of null" on ESLint 3.16.0
When upgrading ESLint to the latest version, this plugin now throws these errors. You can see this for yourself when upgrading the version of ESLint in the `devDependencies` of this package.

```
  1) space-after-type-colon valid type X = (number) => string;:
     TypeError: Cannot read property 'value' of null
      at src/rules/typeColonSpacing/reporter.js:20:9

  2) space-after-type-colon valid type X = (number) => string;:
     TypeError: Cannot read property 'value' of null
      at src/rules/typeColonSpacing/reporter.js:20:9

  3) space-before-type-colon valid type X = (number) => string;:
     TypeError: Cannot read property 'value' of null
      at src/rules/typeColonSpacing/reporter.js:20:9

  4) space-before-type-colon valid type X = (number) => string;:
     TypeError: Cannot read property 'value' of null
      at src/rules/typeColonSpacing/reporter.js:20:9
```

I'll have a PR that fixes this shortly.

wyze:
Something changed between ESLint 3.15.0 and 3.16.0 that caused the token to no longer return. It now returns null, instead of a valid token that had a `value !== ':'`.

This adds a check to make sure that `colon` is not null first.

Could possibly be this refactor: https://github.com/eslint/eslint/commit/834f45d32ed322940dd185a32e5342193fdde914

This fixed #196 as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "chai": "^3.5.0",
     "conventional-changelog-lint": "^1.0.1",
     "create-index": "^0.1.3",
-    "eslint": "^3.4.0",
+    "eslint": "^3.16.0",
     "eslint-config-canonical": "1.8.1",
     "gitdown": "^2.5.0",
     "husky": "^0.11.7",
--- a/src/rules/typeColonSpacing/reporter.js
+++ b/src/rules/typeColonSpacing/reporter.js
@@ -17,7 +17,7 @@ export default (direction, context, {always, allowLineBreak}) => {
     // Support optional names
     // type X = { [string]: a }
     // type X = string => string
-    if (colon.value !== ':') {
+    if (!colon || colon.value !== ':') {
       return;
     }
 


onehorsetown:
Ran into the same problem.  Thanks for the fix!

gajus:
Thank you for this!

masylum:
Did this really make it to the `2.30`? I don't see this change there.

frankthelen:
Release v2.30.0 is dated Jan 15 -- it cannot include this bugfix.
@gajus Anyway, would be fantastic if you create a new release with this fix. Running into this issue too.

gajus:
Should be in the next release.

thealjey:
@gajus If you don't mind me asking. How soon will that be?

RusAlex:
As temporary solution just stick with previous version of `eslint` package.

```json
    "eslint": "~3.12.1",
```

incubus8:
sorry @gajus any idea when this fix will be available? thank you

onehorsetown:
In the interim, simply edit:

```
node_modules/eslint-plugin-flowtype/dist/rules/typeColonSpacing/reporter.js
```

line 36 to:

```
if (!colon || colon.value !== ':') {
```

Not an ideal long term solution, but it will keep you running with less than a dozen keystrokes.

gajus:
Just release 2.30.1.

I didn't see that semantic-release is failing. Sorry.

PR #203, opened 2017-03-07
fix engines

mkwtys:
- https://github.com/eslint/eslint/blob/master/package.json#L119
- https://github.com/babel/babel-eslint/blob/master/package.json#L34

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "semantic-release": "^6.3.2"
   },
   "engines": {
-    "node": ">6.0.0"
+    "node": ">=4"
   },
   "keywords": [
     "eslint",


gajus:
There is no guarantee for v4 support.

hzoo:
What do you mean by no guarantee for v4 support? You are running node 4 in .travis.yml so if that passes then it's supported?

Adding in this https://github.com/gajus/eslint-plugin-flowtype/commit/5de0277871438ea60eb5909e0a4f7a325a3bd4f3 commit you still target node 4 so I'm confused why you decided to add only supporting node 6 and also in a non major version.

STRML:
See also https://github.com/babel/eslint-config-babel/pull/21, which this is causing issues for

gajus:
> What do you mean by no guarantee for v4 support? You are running node 4 in .travis.yml so if that passes then it should be supported?

The plan is to get rid of it at some point in relatively near future.

PR #222, opened 2017-04-18
Feature/no types missing flow annotation

micaste:
I took over from @aaron-harvey for #180 and added one commit to take in consideration the comment from @danharper and support `// @noflow` annotations.


gajus:
Thank you

micaste:
Cool. Thanks for your reactiveness. 

I also thought it would be good to add it to the recommended.js but didn't know what was the policy for the error level (I think it should be 2)

gajus:
> I also thought it would be good to add it to the recommended.js but didn't know what was the policy for the error level (I think it should be 2)

I am not sure about adding it to recommended.

Flow promotes itself as enabling gradual transition into types. This rule would require to make the entire file flow-enabled as soon as one type definition is used.

I've seen quite a few applications start adding types slowly, starting with primitives and more complex types coming into play only in the long play. I can see how this error would become a nuisance.

I'd agree with it throwing a warning, though.

MoOx:
My 2 cents: this rule is missing from the readme :)

micaste:
@gajus that's particulary for projects that are slowly migrating to flow that this rule is useful. It has no use for projects that are already completely typed.

> Flow promotes itself as enabling gradual transition into types. This rule would require to make the entire file flow-enabled as soon as one type definition is used.

The way Flow works, introducing a type in a file without the `@flow` annotation at the top is completely useless. the Flow engine will completely ignore the file. I might be wrong, but I think that enabling Flow for the entire file is a necessity as soon as you introduce one type in the file. Otherwise the type will just be ignored and there is no point writing it...
That's why I would add it to the recommended, as an error (because I think that having some dead typing code is more a threat than anything else).

@MoOx thanks ! https://github.com/gajus/eslint-plugin-flowtype/blob/master/CONTRIBUTING.md states that the README.md is automatically generated and that we shouldn't touch it. I guess that this is not entirely true.

gajus:
> I guess it is refering to /README.md and not /.README/README.md

Yes

> Otherwise the type will just be ignored and there is no point writing it...

Not really.

I've personally did not jump straight into Flow. I've started with https://github.com/codemix/flow-runtime (whatever the predecessor package was).

But... this is "eslint-plugin-flowtype" plugin and not "flow-runtime", therefore your argument makes sense.

Can you please PR `./.README/README.md` and the rule addition?

micaste:
Yes I'll do it. I was waiting for both points to be confirmed.

I had not thought of using flow annotations without Flow itself.

PR #233, opened 2017-05-21
Update `looksLikeFlowFileAnnotation` regex to match only @flow

### LINKED ISSUE #165: require-valid-file-annotation is too aggressive
When configuring `flowtype/require-valid-file-annotation` like so:
```
"flowtype/require-valid-file-annotation": [ "error", "never" ]
```
I am getting the linting errors:
```
1/2 Flow file annotation not at the top of the file. (flowtype/require-valid-file-annotation)
2/2 Malformed Flow file annotation. (flowtype/require-valid-file-annotation)
```
I am experiencing this linting error inside a JS file that contain a bunch of eslint rule configurations that I have annotated with `@fixable` (for rules that are fixable) inside a JSDoc block comment (I appreciate this isn't an official JSDoc type annotation):

flowtype-bug

Looking at the rule implementation, I can see that the [`potentialFlowFileAnnotation`](https://github.com/gajus/eslint-plugin-flowtype/blob/master/src/rules/requireValidFileAnnotation.js#L41) boolean uses [`looksLikeFlowFileAnnotation`](https://github.com/gajus/eslint-plugin-flowtype/blob/master/src/rules/requireValidFileAnnotation.js#L11) to conditionally report errors. 

To my mind, this test is a little aggressive since it just looks for comments starting `@f` (which `@fixable` obviously passes). I see that you have a utility function `fuzzyStringMatch`—could this not be used to increase the certainty that the comment is a `@flow` annotation instead of the `@f` regex?

Robdel12:
The `looksLikeFlowFileAnnotation` regex was too broad and would catch
jsdoc comments like `@fixable` and `@function`, causing eslint errors
in projects that don't use flow (in my case).

I updated the regex to only check for anything beginning with
`@flo`. Is there any reason not to full on match `@flow`? I'm not
familiar with how flow works.

This should fix #165.

### DIFF
--- a/src/rules/requireValidFileAnnotation.js
+++ b/src/rules/requireValidFileAnnotation.js
@@ -9,7 +9,7 @@ const defaults = {
 };
 
 const looksLikeFlowFileAnnotation = (comment) => {
-  return /@(?:no)?f/i.test(comment);
+  return /@(?:no)?flo/i.test(comment);
 };
 
 const isValidAnnotationStyle = (node, style) => {
--- a/tests/rules/assertions/requireValidFileAnnotation.js
+++ b/tests/rules/assertions/requireValidFileAnnotation.js
@@ -168,6 +168,19 @@ export default {
       ]
     },
     {
+      code: '// @function',
+      options: [
+        'never',
+        {
+          annotationStyle: 'none'
+        }
+      ]
+    },
+    {
+      code: '// @fixable',
+      options: [ 'error', 'never' ]
+    },
+    {
       code: '/* @flow */',
       options: [
         'always',
@@ -178,5 +191,3 @@ export default {
     }
   ]
 };
-
-


sunnyvempati:
can we get this merged?

gajus:
I am reluctant to, as I've never run into this situation working on large projects. I'd rather recommend  disabling the rule in effected files. Alternatively, regex could be provided as an option.

Robdel12:
Thankfully this rule was (temporarily) disabled from create react app, so it might make it easier to ignore this. 

I ran into this immediately since I had JSdocs that included `@function` and it triggered a match on this regex. Which caused flow eslint errors when I don't even use flow.

danharper:
👍  for merging this

Robdel12:
Thanks! 😃

gajus:
@danharper Dan has veto rights here. :-)

PR #237, opened 2017-06-05
feat: add prefer-read-only-array (fixes #236)

### LINKED ISSUE #236: Add a rule that requires to use $ReadOnlyArray instead of Array
Rule must warn when `Array` is used instead of [`$ReadOnlyArray`](https://github.com/facebook/flow/blob/v0.46.0/lib/core.js#L185).

gajus:
Adds a rule that requires to use `$ReadOnlyArray` in place of `Array`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,27 +1,30 @@
 import _ from 'lodash';
-import {checkFlowFileAnnotation} from './utilities';
+import {
+  checkFlowFileAnnotation
+} from './utilities';
+import booleanStyle from './rules/booleanStyle';
 import defineFlowType from './rules/defineFlowType';
+import delimiterDangle from './rules/delimiterDangle';
 import genericSpacing from './rules/genericSpacing';
-import noWeakTypes from './rules/noWeakTypes';
+import noDupeKeys from './rules/noDupeKeys';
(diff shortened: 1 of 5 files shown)


gajus:
Need to decide how to implement this:

* If the rule enforces to use `ReadOnlyArray` always, then it should probably be renamed to "no-array"
* or should the rule enforces to use `ReadOnlyArray` only when the array is not modified in the current scope? (To make it consistent with e.g. [`prefer-const`](http://eslint.org/docs/rules/prefer-const)).

I am learning towards having this as a rule that always requires to use `$ReadOnlyArray`. User can disable the rule per-line when mutability is desirable.

@danharper Thoughts about the usefulness of the rule and which way it should be implemented?

src/index.js
```diff
@@ -1,27 +1,30 @@
 import _ from 'lodash';
-import {checkFlowFileAnnotation} from './utilities';
+import {
+  checkFlowFileAnnotation
+} from './utilities';
+import booleanStyle from './rules/booleanStyle';
```
gajus: This is just sorting of the imports in an alphabetical order, as the order has not been kept.

PR #249, opened 2017-06-28
Fixes #248: recognize @flow comments even if it isn't in the first co…

### LINKED ISSUE #248: Files that have a comment before the @flow comment are not considered as being flow-checked
Exemple of such a file:
```javascript
/*
  LICENSE information
*/
// @flow

var i: number = 3;
```

The problem is at https://github.com/gajus/eslint-plugin-flowtype/blob/master/src/utilities/isFlowFile.js#L17

Indeed we forcefully only handle the first comment.

I'm quite sure that flow will flow-check a file wherever the flow comment is, even if their documentation says it should appear before the first line of code. Anyway the current rule doesn't actually check this either :)

julienw:
…mment

### DIFF
--- a/src/utilities/isFlowFile.js
+++ b/src/utilities/isFlowFile.js
@@ -10,11 +10,8 @@ import isFlowFileAnnotation from './isFlowFileAnnotation.js';
 export default (context, strict = true) => {
   const comments = context.getAllComments();
 
-  if (!comments.length) {
-    return false;
+  return comments.some((comment) => {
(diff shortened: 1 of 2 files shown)


scott-silver:
@julienw great minds think alike: https://github.com/gajus/eslint-plugin-flowtype/pull/240

julienw:
Ah indeed :D Now we just need a merge and a new version.

gajus:
Sorry, this slipped my radar. Merged https://github.com/gajus/eslint-plugin-flowtype/pull/240.

Thank you @julienw, @scott-silver.

PR #252, opened 2017-07-09
chore: Rule schemas for validation.

SamirTalwar:
I'm a big fan of schemas for ESLint rules so I don't misconfigure them. This PR adds them to every rule.

There doesn't seem to be a standard way to test these so I added my own ("misconfigured" arrays in the test files). Please let me know if you'd prefer a different way of doing this.

This introduces one bit of weirdness: there's now two test blocks for each rule (and I don't know how to fix this without changing the `RuleTester`). It looks like this:

```
  rule-name
    misconfigured
      …

  rule-name
    invalid
      …
    valid
      …
```

Pretty sure this complies with the contribution guidelines, but if I've missed anything, let me know.


gajus:
Hi, this is a wonderful PR. However, can you please replace `is-my-json-valid` with Ajv please?

Ping @epoberezkin if you need help.

SamirTalwar:
Sure, happy to. I picked *is-my-json-valid* because [ESLint v3 depends on it](https://github.com/eslint/eslint/blob/3.x/package.json). ESLint v4 uses *ajv* though, so it makes sense.

SamirTalwar:
Ah, good, looks like the build passed. Any other requests?

gajus:
> Ah, good, looks like the build passed. Any other requests?

No, prefect. Thank you so much.

SamirTalwar:
Great. Any idea when it'll be published to NPM?

gajus:
Looks like the build failed.

https://travis-ci.org/gajus/eslint-plugin-flowtype/jobs/252330589

I doubt it is anything to do with this PR – just forced a new build https://travis-ci.org/gajus/eslint-plugin-flowtype/builds/252377902.

gajus:
Released.

SamirTalwar:
That build failure is odd, but seems to have fixed itself.

Thanks for releasing!

Arcanemagus:
This PR doesn't actually specify the schema properly according to [the documentation](https://eslint.org/docs/developer-guide/working-with-rules), I'll be putting up a PR fixing this in addition to the documentation URL PR I was working on 😉.

PR #262, opened 2017-07-28
feat: add rule that only requires types for exported functions and variables

AnIrishDuck:
Here at Instructure, we struggled with the right amount of type annotations to add to our code. Adding type annotations to every function, lambda, and variable seemed like overkill. Flow can automatically deduce the vast majority of those types.

On the other hand, the wild west of no annotations could lead to cryptic errors and difficulty understanding how flow arrived at a certain type. Strongly typed languages like Haskell have long held that the right balance is to specifically annotate types on all values that are exported to other modules. We decided we wanted to copy this approach.

This is a first pass at integrating that approach into this plugin. I am not particularly happy with it. Having another way to demand type annotations feels clunky, when there already are `require-variable-type`, `require-parameter-type`, and `require-return-type`.

My idea to fix this was to add an `exportOnly` flag to each rule.

However, when I began refactoring to integrate this approach into those existing rules, they were missing features that I had implemented in `require-export-types`. These features grew out of bolting these annotation requirements onto our existing codebase, and might be controversial:

- not requiring types for simple literals (`export const x = 10`)
- allowing type annotations on LHS and RHS of an expression (`export const v = f(3): number`)
- treating a function with return and parameter type annotations as a fully specified type (i.e. `const f = (x: number, y: number): number => x * y` does not require a type annotation on `f`)

I decided to stop my refactoring effort when these inconsistencies became apparent. I'd like some feedback from the current maintainers before continuing:
- Do you think this feature even belongs in this plugin, or do you think I should create a new eslint plugin?
- Do you think this PR could be included with some minor changes, or would you expect more major refactoring like I described above?

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -151,6 +151,7 @@ When `true`, only checks files with a [`@flow` annotation](http://flowtype.org/d
 {"gitdown": "include", "file": "./rules/no-types-missing-file-annotation.md"}
 {"gitdown": "include", "file": "./rules/no-weak-types.md"}
 {"gitdown": "include", "file": "./rules/object-type-delimiter.md"}
+{"gitdown": "include", "file": "./rules/require-export-type.md"}
 {"gitdown": "include", "file": "./rules/require-parameter-type.md"}
 {"gitdown": "include", "file": "./rules/require-return-type.md"}
 {"gitdown": "include", "file": "./rules/require-valid-file-annotation.md"}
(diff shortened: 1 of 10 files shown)


gajus:
@AnIrishDuck What is the latest regarding this?

gajus:
Please follow up if you would still like to this PR to be merged.

PR #266, opened 2017-08-26
fix: Don't crash on type spread in no-dupe-keys

lydell:
The no-dupe-keys rule used to crash when encountering type spread
(`type A = { ...B }`) with "TypeError: Cannot read property 'type' of
undefined". This commit makes it no longer crash.

For reference, I’ve encountered this in some real world code and had to use this workaround for now:

```js
// Use Flow’s comment syntax here to avoid a problem in eslint-plugin-flowtype:
/* ::
type Two {
  a: number,
  ...One
}
*/
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "devDependencies": {
     "ajv": "^5.2.1",
     "babel-cli": "^6.14.0",
-    "babel-eslint": "^6.1.2",
+    "babel-eslint": "^7.2.3",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-transform-object-rest-spread": "^6.23.0",
     "babel-preset-env": "^1.1.10",
--- a/src/rules/defineFlowType.js
+++ b/src/rules/defineFlowType.js
@@ -49,9 +49,6 @@ const create = (context) => {
         makeDefined(qid);
       }
     },
-    InterfaceDeclaration (node) {
-      makeDefined(node.id);
-    },
     Program () {
       globalScope = context.getScope();
     },
--- a/src/rules/noDupeKeys.js
+++ b/src/rules/noDupeKeys.js
@@ -61,6 +61,10 @@ const create = (context) => {
     const haystack = [];
 
     _.forEach(node.properties, (identifierNode) => {
+      if (identifierNode.type === 'ObjectTypeSpreadProperty') {
+        return;
+      }
+
       const needle = {name: getParameterName(identifierNode, context)};
 
       if (identifierNode.value.type === 'FunctionTypeAnnotation') {
--- a/src/rules/useFlowType.js
+++ b/src/rules/useFlowType.js
@@ -9,29 +9,7 @@ const create = (context) => {
     DeclareClass: markTypeAsUsed,
     DeclareFunction: markTypeAsUsed,
     DeclareModule: markTypeAsUsed,
-    DeclareVariable: markTypeAsUsed,
-    GenericTypeAnnotation (node) {
-      let typeId;
-      let scope;
-      let variable;
-
-      if (node.id.type === 'Identifier') {
-        typeId = node.id;
-      } else if (node.id.type === 'QualifiedTypeIdentifier') {
-        typeId = node.id;
-        do {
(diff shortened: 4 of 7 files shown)


lydell:
I had to update babel-eslint from 6 to 7 for the type spread. Unfortunately, that caused some use-flow-type and define-flow-type tests to break.

### use-flow-type

These tests:

https://github.com/gajus/eslint-plugin-flowtype/blob/6eceb89551700701bad8a9d9ab5cab084f3a44fe/tests/rules/assertions/useFlowType.js#L38-L93

Those cases are expected to be invalid, but I can't understand why. To me they look valid. So I'm unsure what to do here! Some  help would be nice.

### define-flow-type

The tests failing here look legit, and I'll try to fix them.

lydell:
Progress: I’ve `git bisect`:ed babel-eslint, and this is the first “bad” commit: https://github.com/babel/babel-eslint/commit/a2c3b30e6a86a39e0cb44a95a1990ca54415bd1f

lydell:
Ahh! babel-eslint now intentionally makes sure that no-unused-vars does not report anything for these cases: https://github.com/babel/babel-eslint/pull/444

So I guess I can just remove the tests?

lydell:
For reference, the PR which handles no-undef for interface declarations is here: https://github.com/babel/babel-eslint/pull/449

I’ve now moved the failing tests from `VALID_WITH_USE_FLOW_TYPE` to `ALWAYS_VALID` so that the tests pass again.

I also removed the now dead code from the use-flow-type rule and the define-flow-type rule, but I just realized that by doing so we lose babel-eslint@6 compatibility (actually, only babel-eslint@^7.2.1 works). Do you want me to restore that code to maintain compatibility with older babel-eslint versions?

lydell:
@gajus or anyone:

There are 3 ways this PR can be done:

### 1

Add these 3 lines:

```js
if (identifierNode.type === 'ObjectTypeSpreadProperty') {
  return;
}
```

And this test (commented out on purpose):

```js
// TODO: Uncomment when we update the tests to use babel-eslint@^7.2.0.
// {
//   code: 'type One = { c: number }; type Two = { a: number, b: string, ...One }'
// }
```

### 2

Same as above, but also update the tests to use `babel-eslint@^7.2.0` so the new test does not have to be commented out. This also requires moving some tests from `VALID_WITH_USE_FLOW_TYPE` to `ALWAYS_VALID`, but it should be fine.

### 3

Same as above, but also remove dead code since `babel-eslint@^7.1.0` or so.

---

This PR is currently at alternative **3** – are you happy with this, or would you like me to change to alternative 1 or 2 to get this merged?

Thanks! :)

gajus:
It is fine as is, just waiting for Babel v7 to be released.

lydell:
Ok! 👍 

Curious question: What has Babel v7 to do with this? :)

gajus:
`babel-eslint` depends on Babel core libs, which are in alpha at the moment.

https://github.com/babel/babel-eslint/blob/42d0c5b20cf65ae4100d06aa8dcaf372723f378d/package.json#L15-L18

lydell:
@gajus Since it is not known when Babel v7 goes out of alpha, would you have anything against adding the following three lines without tests:

```js
if (identifierNode.type === 'ObjectTypeSpreadProperty') {
  return;
}
```

This way we could do a patch release immediately so people can use type spreads without hassle. When Babel v7 goes out of alpha, we can merge this PR.

gajus:
Babel 7 is not far from getting released. Days, maybe weeks.

https://github.com/babel/babel/milestone/9

I would recommend simply disabling the `no-dupe-keys` in the meantime.

lydell:
Looks like I need to merge master again ... and now even more tests need to be updated because of babel-eslint. I'll try to do this soon.

kusmierz:
@lydell any news here? :)

lydell:
Yes: I never got around merging master again. And I'm not particularly keen on doing it.

lydell:
Yes: I never got around merging master again. And I'm not particularly keen on doing it.

lydell:
I don’t think this PR is relevant anymore.

PR #272, opened 2017-09-16
feat: add rule no-unused-expressions

mkscrg:
This rule extends the ESLint's `no-unused-expressions` to ignore type cast expressions.

ESLint's `no-unused-expressions` catches some likely mistakes, but it also complains about bare type cast expressions. These are unlikely to be mistakes and are useful in several situations. See `.README/rules/no-unused-expressions.md` for an example.


mkscrg:
@gajus what do you think?

gajus:
Thank you

pronebird:
How can I enable `no-unused-expressions` for Flow casts, but still throw errors for unused expressions in Javascript?

```
"no-unused-expressions": "error",
```

Simply throws errors for everything.

Issue #212, opened 2017-03-29
Make flowtype/sort-keys fixable


jpolo:
Hi,

It would be nice to make the rule named `flowtype/sort-keys` auto-fixable. Differently from the object key sorting, it should not be as difficult because there is no side effects related.

Thanks

deecewan:
just started working on this. @gajus I can't actually see where the rules are specified as `fixable`. For example, in the `booleanStyle.js` rule, there is no mention of `fixable`, however the README mentions that it is in fact fixable.

Could you point me in the right direction?

gajus:
The `fix` method, e.g. https://github.com/gajus/eslint-plugin-flowtype/blob/6eceb89551700701bad8a9d9ab5cab084f3a44fe/src/rules/booleanStyle.js#L17-L19

gajus:
This might come handy too https://github.com/eslint/eslint/pull/7715

deecewan:
@gajus i figured I can just add the `fix` method, without updating the metadata of the rule - i imagine this is due to the older version of ESLint in use?

gajus:
> @gajus i figured I can just add the fix method, without updating the metadata of the rule - i imagine this is due to the older version of ESLint in use?

I am not familiar with the API changes in the recent version of the ESLint.

deecewan:
From [here](https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes), they mention that 

> Unless the rule exports the `meta.fixable` property, ESLint does not apply fixes even if the rule implements fix functions.

gajus:
Might be a good cause for another PR. :-)

PR #273, opened 2017-09-19
add an auto-fix option to sortKeys

deecewan:
This adds an option to `--fix` the `sort-keys` rule.

It would be much cleaner using ESLint v4, because multiple tokens can be replaced. However, this is a fairly good (imo) solution, and it tries hard to maintain code style.

edit: ~this currently does not work in a few cases. Namely, `a: (param: number) => void` (or any function type). It also *doesn't* maintain the `?` in `a?: Object`.~

Alright, I think that covers all the use-cases now. If you can think of any more, please let me know.

Fixes: https://github.com/gajus/eslint-plugin-flowtype/issues/212

### DIFF
--- a/README.md
+++ b/README.md
@@ -2081,6 +2081,8 @@ type FooType = {}
 <a name="eslint-plugin-flowtype-rules-sort-keys"></a>
 ### <code>sort-keys</code>
 
+_The `--fix` option on the command line automatically fixes problems reported by this rule._
+
 Enforces sorting of Object annotations.
 
 This rule mirrors ESlint's [sort-keys](http://eslint.org/docs/rules/sort-keys) rule.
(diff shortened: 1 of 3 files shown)


deecewan:
@gajus any feedback on this?

gajus:
That was not intentional. For whatever reason Git is not giving me an option revert this PR.

gajus:
I have added to commits that show that this PR does not work with covariant and contravariant properties:

* a439e363654286143bd6966620e8d83dc3c9ea94
* 13670d880fff7bf16c1f0d61b2ca7ee28057766b

Can you pick up the latest code from master and follow up?

gajus:
Furthermore, need to check if nested properties work, e.g.

```js
type FooType = {
  a: number,
  c: {
    a: number,
    c: number,
    b: string,
  },
  b: string,
}

```

deecewan:
I'll take a look in the morning

PR #274, opened 2017-09-23
feat: Implement define-flow-type for declare type

jez:
We weren't covering the `declare` versions of Flow type/interface declarations:

    declare type Foo = 'foo';
    declare interface Foo {};


jez:
Rebased on `master`. Would love if I could get some 👀 on this!

Issue #94, opened 2016-08-30
Add rule to enforce one array notation style


opatut:
Instead of `Array<string>` one can write `string[]`.

I would like to see a rule that enforces either style. 

Bonus points for an option that only enables the rule on "simple" types, such as `TFoo` or `string` (i.e. enforce `string[]` instead of `Array<string>`) but not on more complex types, e.g. `?TFoo` (compare `Array<?TFoo>` and `(?TFoo)[]`) or `{id: number}` (compare `({id: number})[]` and `Array<{id: number}>`). PS: I would count `Promise<Foo>` and other polymorphic types as simple types, so `Promise<Foo>[]` is legal.

danharper:
I guess this too `?(string[])`

opatut:
So basically we have four rules:

* require-simple-array-shorthand
* no-simple-array-shorthand
* require-complex-array-shorthand
* no-complex-array-shorthand

That way, one can decide to enforce either style for simple types, but not enforce them for complex styles. Or vice versa. Or enforce different styles for different complexities. Or the same for both. Or none at all. Wow. This is complicated.

pnevyk:
@opatut I created a PR (#278) but I need opinions on some points. It would be great if you (or anyone else interested in this feature) leave a comment there.

gajus:
Fixed.

PR #278, opened 2017-09-29
feat: Array style rules

pnevyk:
This pull request adds two new rules which handle functionality requested in #94. I have few points which I need to resolve before making it into the finished state. Also I am going to implement `fix` functionality.

### Default values

Default value for complex types is `Array<Type>` and for simple types is `Type[]`. I personally think that this will be the most common case, but on the other side it brings some type of default inconsistency into the codebase.

### Rest parameters in functions

Array notation of rest parameters in function corresponds to notation enforced by these rules. I was thinking if forcing `Array<Type>` notation in this particular would be better.

### Complex/simple type

Are my decisions in these cases correct?

* Tuples (complex)
* Utility types such as `$Keys<Type>` (simple)
* Array type in shorthand notation (simple)

### Error messages

At this moment, error messages are quite primitive. I am going to improve them. However, I think that just copying the raw string of array element type into the message is not wanted behaviour because in case of multiline type definitions there would be newlines in messages and that could cause problems in linter UIs. I am thinking about removing newlines and test the type for length, shorter ones will be written into message, longer ones will be replaced with a substitution (`Type`).


gajus:
I personally only use `Array<T>` notion. Never used `string[]`.

To be precise, 99% of the time I will be using `$ReadOnlyArray<T>`. It is very rare I'd use `Array<T>`.

pnevyk:
After rethinking that, I agree with you in shorthand usage. I suppose I should change default config for simple types to `verbose`.

Also, thank you for reminding me the `$ReadOnlyArray<T>` type. I should definitely support this one in the rules.

pnevyk:
I think the rules are finished now. I suppose I should squash all commits into one. I will do it as soon as I get the approval.

gajus:
Whats the state of this PR?

pnevyk:
I hoped that I will get some feedback, for example from ones who discuss this feature in #94. From my point of view, it is ready to merge, if you agree after review of my code.

gajus:
The original description says:

> I have few points which I need to resolve before making it into the finished state. Also I am going to implement fix functionality.

Is this still relevant?

pnevyk:
The things I had in mind were these specified in the description, so:
* **default values** - I agreed with you that `Array<T>` should be default for both cases
* **rest parameters** - it was just an idea, it would probably make the code more complex and there is no strong reason to have it
* **complex/simple type** - I listed cases where I was not 100% sure if I put those types into correct category so I thought that somebody either ensure me or give me his opinion why a type should be in the other category
* **error messages** - I show inlined types in the error message if they don't exceed 50 characters, again, it was my way of thinking but there may be a better one
* **fix** - fix functionality is implemented

pnevyk:
Am I supposed to do anything here yet?

kusmierz:
@gajus, any chance to merge it? :) It looks very nice though.

micmcg:
Would love to see this merged!

lukeapage:
Also would love to see this merged @gajus

gajus:
Thank you @pnevyk

PR #279, opened 2017-10-01
feat: Add `require-read-only-array` rule (#236)

pnevyk:
This rule enforces use of `$ReadOnlyArray` instead of `Array` or array
shorthand notation.


pnevyk:
Well, right after I submitted this PR I noticed there is already one solving this issue. I am so sorry for this duplicate, it should be probably closed. Next time, I will pay more attention.

gajus:
@pnevyk If you do not mind, I prefer your PR.

pnevyk:
Ok, that's fine. And is there already a decision in points you noted in your PR (name `no-array` and detection of mutation in scope)?

gajus:
> detection of mutation in scope

No need. If user intends to perform a mutation, he will just suppress the warning. Otherwise, it is easy to perform accidental mutation without even being aware of this rule.

As for the rule name, "no-array" is a bit too generic. I think "no-mutable-array" would work the best.

pnevyk:
@gajus So I renamed the rule to `no-mutable-array`. Is there anything to be done yet?

gajus:
Thank you

PR #280, opened 2017-10-04
Fix some issues with the sortKeys autoFix

deecewan:
The fix was accidentally merged (#273) while there were still some issues with it.

The outstanding issues have now been fixed - if there are other things I've overlooked, please let me know.

I'm not convinced on the recursive structure, but it's the only way (I know of) to apply the nested fixes - otherwise they'll just continually conflict.

cc @gajus

### DIFF
--- a/.README/rules/sort-keys.md
+++ b/.README/rules/sort-keys.md
@@ -1,5 +1,7 @@
 ### `sort-keys`
 
+_The `--fix` option on the command line automatically fixes problems reported by this rule._
+
 Enforces sorting of Object annotations.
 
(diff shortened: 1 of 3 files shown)


gajus:
> I'm not convinced on the recursive structure, but it's the only way (I know of) to apply the nested fixes - otherwise they'll just continually conflict.

Any recursive structure can be converted to a loop. Somewhat relevant: https://stackoverflow.com/questions/159590/way-to-go-from-recursion-to-iteration.

Anyway, I don't see it as a problem.

PR #283, opened 2017-11-02
fix: Mark opaque type identifiers as defined with define-flow-type

### LINKED ISSUE #260: Support new Flow "opaque type" syntax
https://flow.org/en/docs/types/opaque-types/

Right now this new Flow 0.51 syntax produces eslint errors.

valscion:
This PR fixes #260 by suppressing invalid `no-undef` warnings when `define-flow-type` is active.

It seems like there's some upstream problems with opaque types causing `opaque type A = Array` to report `A` as `no-undef`, even though it shouldn't. Does anyone happen to have an idea as to which piece of this dependency graph would be responsible for fixing that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,9 @@ var a: AType<BType>
 type A = AType
 // Additional rules: {"no-undef":2}
 
+opaque type A = AType
+// Additional rules: {"no-undef":2}
+
 function f(a: AType) {}
 // Additional rules: {"no-undef":2}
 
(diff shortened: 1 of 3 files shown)


gajus:
Whats the state of this PR?

valscion:
> Whats the state of this PR?

I rebased this on top of `master` and verified on my own codebase that this fixes the `no-undef` that gets raised with `opaque type A = Array` code.

```js
// @flow

opaque type A = Array;
```

## Before, with `v2.30.1`

screen shot 2017-12-11 at 10 09 36

## After, with this PR

screen shot 2017-12-11 at 10 10 08

---

The cases raising `no-undef` error without `define-flow-type` rule active are mysterious to me, and are likely problems with some upstream dependency. When `define-flow-type` is active, that possible upstream bug also disappears.

All in all, I think that this PR is ready to be merged and shipped — there's nothing extra that needs to be done here.

PR #298, opened 2018-01-06
feat: add no-flow-fix-me-comments

TheSavior:
We use this lint rule internally at Facebook to make sure we keep an eye on these flow suppression comments and ensure they get fixed over time. We figure this might be useful to others as well.

cc @bestander


bestander:
Nice, thanks for doing it, @TheSavior!

Issue #293, opened 2017-12-05
error Cannot read property 'type' of undefined


StevePotter:
Hi, we are using version 2.39.1 and I suddenly got an error when creating a Flow type from spreading other types.  This is a feature of Flow that allows union types for sealed objects.  Without the spread syntax, I cannot get Flow to work right.  So Flow worked but then lint failed.

It fails on all of the eslint-flowtype rules.  Only if I remove them from .eslintrc.json does it work.  Disabling all the rules in a /* eslint-disable */ did not fix it.

The Flow type that's causing it:
```
type DataProps = {
  ...ApolloQueryResult,
  ...ClaimsQuery,
  refetch: () => Promise<Object>,
  loadMoreClaims: () => Promise<Object>,
}
```

Stack trace:
```
Cannot read property 'type' of undefined
TypeError: Cannot read property 'type' of undefined
    at /private/tmp/mobile/node_modules/eslint-plugin-flowtype/dist/rules/noDupeKeys.js:76:31
    at arrayEach (/private/tmp/mobile/node_modules/lodash/lodash.js:537:11)
    at Function.forEach (/private/tmp/mobile/node_modules/lodash/lodash.js:9359:14)
    at checkForDuplicates (/private/tmp/mobile/node_modules/eslint-plugin-flowtype/dist/rules/noDupeKeys.js:73:22)
    at listeners.(anonymous function).forEach.listener (/private/tmp/mobile/node_modules/eslint/lib/util/safe-emitter.js:47:58)
    at Array.forEach (native)
    at Object.emit (/private/tmp/mobile/node_modules/eslint/lib/util/safe-emitter.js:47:38)
    at NodeEventGenerator.applySelector (/private/tmp/mobile/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/private/tmp/mobile/node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (/private/tmp/mobile/node_modules/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (/private/tmp/mobile/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:608:23)
```

I assume this is a Flow feature that hasn't been taken into account.  I'd be happy to work on a fix.  Any tips for getting started?

valentinvoilean:
Any news about this issue ?

[...]

PR #302, opened 2018-01-18
Prevent crash when using types with spreads

StevePotter:
Fixes https://github.com/gajus/eslint-plugin-flowtype/issues/293

This just avoids a crash.  It doesn't go any further.  But at least the plugin is usable :)

Oh man, I just saw https://github.com/gajus/eslint-plugin-flowtype/pull/266.  This might be a duplicate although that PR seems more ambitious so I'll leave this PR open

### DIFF
--- a/src/rules/noDupeKeys.js
+++ b/src/rules/noDupeKeys.js
@@ -62,6 +62,10 @@ const create = (context) => {
 
     _.forEach(node.properties, (identifierNode) => {
       const needle = {name: getParameterName(identifierNode, context)};
+      // avoids error for types constructed from spreading others like: type Props = { ...A, ...B }
+      if (identifierNode.type === 'ObjectTypeSpreadProperty') {
+        return
+      }
 
       if (identifierNode.value.type === 'FunctionTypeAnnotation') {
         needle.args = _.map(identifierNode.value.params, (param) => {


src/rules/noDupeKeys.js
```diff
@@ -62,6 +62,10 @@ const create = (context) => {
 
     _.forEach(node.properties, (identifierNode) => {
       const needle = {name: getParameterName(identifierNode, context)};
+      // avoids error for types constructed from spreading others like: type Props = { ...A, ...B }
+      if (identifierNode.type === 'ObjectTypeSpreadProperty') {
+        return
```
serchavalos: missing `;` bro 😄

suchipi:
@gajus I know babel 7 is still beta, but is there any chance this could be merged before its release? This PR is not very intrusive.

MoOx:
This is not tied to babel 7. You can use latest babel-eslint (8.x) with babel 6 and get this problem. This PR fix the bug I have. Would be nice to see this merged!

@StevePotter could you make CI green? There is some easy fix here https://travis-ci.org/gajus/eslint-plugin-flowtype/jobs/330540986#L2043

kusmierz:
The same issue seems to appear for `sortKeys` rule:
```
Cannot read property 'type' of undefined
TypeError: Cannot read property 'type' of undefined
    at ./node_modules/eslint-plugin-flowtype/dist/rules/sortKeys.js:85:24
```

IanVS:
https://github.com/gajus/eslint-plugin-flowtype/issues/293 is closed/fixed, so I think this PR can be closed as well now?  Although I don't know if `sortKeys` is still broken (I don't use it).

gajus:
> Although I don't know if sortKeys is still broken (I don't use it).

That would be a separate issue anyway.

PR #305, opened 2018-01-24
feat: add require-exact-type rule (fixes #304)

### LINKED ISSUE #304: Enforce exact object types
It would be interesting and useful to be able to enforce the use of exact object types for type definitions (see [here](https://flow.org/en/docs/types/objects/#toc-exact-object-types)).

Example:

```ts
// Don't allow the following
type StateType = { query: string };
const state: StateType = { query: 'hello', foo: 'bar' }; // valid

// Enforce the following
type StateType = {| query: string |};
// or
type StateType = $Exact<{ query: string }>;

const state: StateType = { query: 'hello', foo: 'bar' }; // invalid
const state: StateType = { query: 'hello' }; // valid
```

grit96:
Adds support for enforcing exact object types.


gajus:
Thank you

Issue #290, opened 2017-11-21
Eslint doesnt know about $ReadOnly utility type


smdern:
`$ReadOnly` was added in flow `0.59`

https://github.com/facebook/flow/blob/master/Changelog.md#0590

gajus:
PR welcome.

lll000111:
I don't get an error. Demo code &mdash; both Flow and eslint report no errors:

```js
// @flow
'use strict';

type MyObj = {| foo: number |};

/**
 * @param obj
 */
function demo (obj: $ReadOnly<MyObj>): void {

}

demo({foo: 42});
```

mattdarveniza:
I also get an eslint no-undef violation when any flow utility type is used. I don't see any reference to these global utility types anywhere in the repository, so I'm a little confused as to how eslint _wouldn't_ raise a no-undef violation for those that can't reproduce.

The above code produces the following error output for me:

```
error  '$ReadOnly' is not defined                    no-undef
```

The referenced PR would resolve this.

n1ru4l:
I can confirm this issue by using the following versions

```
    "eslint-plugin-flowtype": "2.46.1",
    "eslint-plugin-jest": "21.15.0",
    "eslint-plugin-prettier": "2.6.0",
    "babel-core": "6.26.0",
    "babel-eslint": "8.2.2",
    "babel-plugin-macros": "2.2.0",
    "babel-plugin-transform-export-extensions": "6.22.0",
    "babel-plugin-transform-flow-strip-types": "6.22.0",
    "babel-plugin-transform-object-rest-spread": "6.26.0",
    "babel-plugin-transform-runtime": "6.23.0",
    "babel-preset-env": "1.6.1",
```

**.eslintrc.js**
```js
`use strict`

module.exports = {
  root: true,
  plugins: [`prettier`, `jest`, `flowtype`],
  parser: `babel-eslint`,
  parserOptions: {
    ecmaVersion: 8,
    sourceType: `module`,
    ecmaFeatures: {
      jsx: true,
      experimentalObjectRestSpread: true,
    },
  },
  env: {
    node: true,
    es6: true,
  },
  extends: [`eslint:recommended`, `plugin:jest/recommended`],
  rules: {
    'prefer-const': [`error`],
    'no-var': [`error`],
  },
  overrides: [
    {
      files: ['src/**/__tests__/**/*.js'],
      env: {
        'jest/globals': true,
      },
    },
  ],
}

```

[...]

PR #308, opened 2018-02-01
Add utility types as globals

dpeek:
I ran into #290 so took a naive stab at fixing it. Didn't seem like other utility types were being handled either, so I added them as globals.

Right now you opt-in by putting this in your config:

```
"env": {
    "flowtype/globals": true
},
```

Let me know if I've put it in the wrong place and I'll fix up. Also not sure if you can only specify globals when a file has `@flow` marker...


oliverodaa:
This seems like a great idea addition but it looks like it has not been merged yet. Why not?

gajus:
> flowtype/globals

Where is this documented?

gajus:
Cannot replicate the issue.

Please correct me if I am wrong and will re-open.

PR #317, opened 2018-02-15
fix: object-type-delimeter with new babel-eslint

mwiencek:
Using babel-eslint 8.2.1, the object-type-delimeter rule does nothing, apparently because the `ObjectType{CallProperty,Indexer,Property}` nodes no longer contain trailing semicolons/commas, but their parent `ObjectTypeAnnotation` nodes do.

I could not find a changelog or git commit entry in any of the associated projects for where/why this was changed...

I tried to keep compatibility with both cases, because it's not clear what version this project is targeting. However, I don't know of a good way to test the new behavior given the version of babel-eslint now locked in devDependencies.

### DIFF
--- a/src/rules/objectTypeDelimiter.js
+++ b/src/rules/objectTypeDelimiter.js
@@ -24,8 +24,18 @@ const create = (context) => {
   }
 
   const requireProperPunctuation = (node) => {
-    const tokens = context.getSourceCode().getTokens(node);
-    const lastToken = tokens[tokens.length - 1];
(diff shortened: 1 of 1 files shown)


mwiencek:
I did test the new code path in my own project by manually updating node_modules/eslint-plugin-flowtype/dist/rules/objectTypeDelimiter.js with these changes, and running eslint; the object-type-delimeter rule then started reporting errors as expected. Since tests pass here, it appears to still work with the older dependency too.

Issue #285, opened 2017-11-02
[Request]: Require Type Definitions First


bradennapier:
I can see it being a good rule to enforce all type definitions being at the top of the file. 

**good**
```js
/* @flow */

type MyType = number

const b: MyType = 2

function foo(): MyType {
     return b
}
```

**bad**

```js
/* @flow */

function foo(): MyType {
     return b
}

type MyType = number 

const b: MyType = 2
```

pnevyk:
Related to #168

Issue #168, opened 2016-12-05
Require types to be declared at the top of the file (after import)


gajus:
**Please describe what the rule should do:**

Require that all `type` declarations appear at the top of the file, first thing after any import declarations.

**What category of rule is this? (place an "X" next to just one item)**

[x] Enforces code style
[ ] Warns about a potential error
[ ] Suggests an alternate way of doing something
[ ] Other (please specify:)

**Provide 2-3 code examples that this rule will warn about:**

Good:

```js
import createAsyncTracker from '../createAsyncTracker';
import type {
  AdType
} from '../types';

type Unit = {|
  prefix: string,
  ignoreSuffix?: Array<string>,
  size: string,
  id: string,
  max?: number
|};

```

Bad:

```js
import createAsyncTracker from '../createAsyncTracker';
import type {
  AdType
} from '../types';

const foo = 'bar';

type Unit = {|
  prefix: string,
  ignoreSuffix?: Array<string>,
  size: string,
  id: string,
  max?: number
|};

```

gajus:
@danharper Might be a good idea to create an issue template, like ESLint does:

* http://eslint.org/docs/developer-guide/contributing/new-rules
* https://raw.githubusercontent.com/eslint/eslint/master/templates/rule-proposal.md

Do you want me to create one?

Is the above information sufficient? (This issue)

danharper:
👍  on an issue template

PR #319, opened 2018-02-18
feat: Add require-types-at-top rule

pnevyk:
Implements rule requested in #168, #285.

The order of imports and types is not enforced, i.e. having import declarations after type declarations is valid, because I think this should be handled in another rule.


gajus:
Thank you

PR #320, opened 2018-02-21
feat: Add type-import-style rule

danwang:
Adds a lint rule that discriminates between two import styles:

```
import type {T, U, V} from 'foo';

import {type T, type U, type V} from 'foo';
```

The latter is superior because it's easier to edit/refactor (adding non-type import entries doesn't require changing the entire line import statement).


gajus:
Thank you.

PR #332, opened 2018-05-21
feat: add fixer for requireValidFileAnnotation when always

KevinGrandon:
If folks add 'always' to their eslint config we can infer that they generally want flow coverage in their codebases. As such, we should be able to autofix files by adding the annotation to make their lives easier.


gajus:
If file starts with `/* eslint-disable require-valid-annotation-flow-autofix */`, then the fix should not be applied. Otherwise, there will be no way to suppress this fix for specific files.

KevinGrandon:
> If file starts with /* eslint-disable require-valid-annotation-flow-autofix */, then the fix should not be applied. Otherwise, there will be no way to suppress this fix for specific files.

Thank you for chiming in and sorry for the slow follow-up. Would you happen to have an example of an eslint plugin that does this?

I think that another way to prevent autofixing of files would be to add a // @noflow file declaration to the top of the file? Happy to add some documentation if that would be sufficient?

gajus:
> I think that another way to prevent autofixing of files would be to add a // @noflow file declaration to the top of the file?

Sounds reasonable.

KevinGrandon:
@gajus - Added a note in the README about autofixing, and how to avoid autofixing this rule per-file. Is there anything else you were looking for here?

gajus:
Thank you

PR #350, opened 2018-08-24
feat: add 'prefer-declaration' option for type-import-style

elCarda:
Introduce new option `prefer-declaration` for type-imports-style rule.

Work the same as `declaration` option but allows the mixture of type and non-type imports.

```javascript
import {x, type B} from "x"; // Valid

import {type A, type B} from "x"; // Not Valid 
import type {A, B} from "x"; // Valid

```

**Reason:**
We found a problem in different behavior of types imports and I would like to add a support for it. 

[Long story](https://github.com/babel/babel/issues/6300#issuecomment-331787868) short:
```javascript
import {type A} from "x";

// is an equivalent of 
import "x";
import type {A} from "x";
```
The first import can contains side effects or can cause circular dependencies and that's why we would like to change/add behavior to prefer `import type {AType} from '';` whenever it is possible but don't loose the ability to use syntactic sugar with mixture of type and non-type imports (`import {AClass, type AType} from ''`).

### DIFF
--- a/.README/rules/type-import-style.md
+++ b/.README/rules/type-import-style.md
@@ -10,6 +10,9 @@ import {type T, type U, type V} from '...';
 
 // 'declaration' style
 import type {T, U, V} from '...';
+
+// 'mixed' style
+import {T, type U, type V} from '...';
 ```
 
 The rule has a string option:
@@ -17,6 +20,8 @@ The rule has a string option:
 * `"identifier"` (default): Enforces that type imports are all in the
   'identifier' style.
 * `"declaration"`: Enforces that type imports are all in the 'declaration'
-  style.
+* `"prefer-declaration"`: Less restrictive than 'declaration'. Allow mixture of types and non-types imports. But if all
+  imported are types the 'declaration' style is allowed only.
+
 
 <!-- assertions typeImportStyle -->
--- a/src/rules/typeImportStyle.js
+++ b/src/rules/typeImportStyle.js
@@ -1,6 +1,6 @@
 const schema = [
   {
-    enum: ['declaration', 'identifier'],
+    enum: ['declaration', 'identifier', 'prefer-declaration'],
     type: 'string'
   }
 ];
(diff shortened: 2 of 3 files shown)


elCarda:
With delay of few days. I think that I should probably introduce a new configuration option for this use case, because this can be a breaking change for people who wants to avoid mixtures of type vs non-type imports... WDYT @gajus, @danwang?

gajus:
> I think that I should probably introduce a new configuration option for this use case

Happy to merge if this is turned into a configuration.

elCarda:
Please wait with merging. I'm trying multiple supported way of imports and already found an issue with `import a, {type B} from 'x';`... Will fix it soon.

elCarda:
@gajus Ready to review.

elCarda:
Well yes we would love to use it. I think that the unpredictability comes form the babel/flow libraries and fact that `import type {A} from 'a';` is not equal to `import {type A} from 'a';` and that makes in our eyes the `identifier` option very dangerous. So it can be taken as deny identifiers only type imports.

I'm not super sure about the name of the added option, better name may help make it more predictable. Any idea?

It's up to you, we temporary switched to `dictionary` which implied a lot of manual rearrangement to split combined imports in our code base, but its already done. We would like to use the syntactic sugar to reduce number of imports for common patter `import {Compontent, type PropType as ComponentPropType} from '../component';` but heaving it at two lines is not something what will ruin our productivity.

gajus:
I would go as far as to suggest that we should remove "identifier" option altogether. As you've mentioned, it introduces side-effects.

It is a shame that as a ESLint plugin author I don't have a way to collect information about how popular different configurations of the plugin are... This type of insight would allow to prioritise development on real usage, rather than the whim of the maintainers.

If you have already refactored to use "declaration", then I would like to keep the current PR open to establish whether there is a demand from the broader audience for such a rule change. 👍/ 👎

PR #355, opened 2018-09-21
Modify require-return-type about annotateUndefined

HsuTing:
With the following:
```js
async () => {};
```
This code will throw error when the setting is `['always', { 'annotateUndefined': false }]`. The `async function` is return `Promise<void>`, but I think this is the same as `void` for `async function`. So, this PR is used to make `Promise<void>` like `void` when using `async function`.


HsuTing:
Hello, will this problem be solved?

gajus:
Thank you

Issue #336, opened 2018-06-01
Broken tests with updated eslint and babel-eslint


pnevyk:
Currently, when `eslint` is updated to `4.19.0` and `babel-eslint` to `8.2.3`, few tests of the plugin are broken. Here is the list:

* **define-flow-type** - Codes containing `interface ...` should throw `no-undef` errors but don't.
* **no-flow-fix-me-comments** - No errors thrown. #307 is related issue.
* **object-type-delimiter** - This seems to be fixed in #317 .
* **sort-keys** - Some variance tests broken. Note that despite that #334 relates to variances in `sort-keys`, the snippet in that issue is fixed incorrectly even with older `eslint` and `eslint-babel`, so it is a separate problem.
* **use-flow-type** - Some `no-unused-vars` tests broken.

These failing tests seem to be blocker for #318.

I will try to look at some of those, but any help is appreciated.

gajus:
> I will try to look at some of those, but any help is appreciated.

I've tried updating those as well but steered away as I've realised it is not a quick fix.

Thank you for investigating.

AodhanHayter:
Good to know others are seeing this as well. I'm getting false positives for `no-unused-vars` concerning imported types, for example:

```javascript
import type Foo from './foo.js' // reported as unused

class Thing {
  constructor(foo: Foo) {
    this.bar = foo.bar
  }
}
```
* `eslint@4.19.1`
* `babel-eslint@8.2.3`
* 'eslint-plugin-flowtype2.49.0`

This problem is remedied in my case by reverting to `babel-eslint@8.0.3`

pnevyk:
So the problem is probably that babel-eslint started to use eslint's new scope manager API ([here](https://github.com/babel/babel-eslint/blob/master/lib/analyze-scope.js)) where a custom Referencer, which visits also type annotations, is implemented. This change was introduced in babel-eslint@8.1.0. This is the reason why our are failing.

This new behavior kind of collides with eslint-plugin-flowtype's `no-undef` and `no-unused-vars` fixes. @gajus Do you have any idea what we should do? Perhaps we can let babel-eslint do the scope tasks eventually.

pnevyk:
I got `no-flow-fix-me-comments` and `sort-keys` working. So the only thing missing is those `no-undef` and `no-unused-vars` errors.

gajus:
> @gajus Do you have any idea what we should do? Perhaps we can let babel-eslint do the scope tasks eventually.

[...]

PR #359, opened 2018-09-27
WIP: Bump versions of babel-eslint, eslint and eslint-config-canonical

pnevyk:
I bumped versions of `babel-eslint`, `eslint` and `eslint-config-canonical`. Because `babel-eslint` have changed their scope analysis in [this](https://github.com/babel/babel-eslint/commit/dbc6546e0713a83c3362bb28d2d4604dbbebcf19) commit to visit also *some* type annotations, some tests were failing (related issue: #336), in particular, `define-flow-type` and `use-flow-type`.

Tests in `use-flow-type` rule were failing because the `no-unused` lint error wasn't being thrown anymore. Tests in `define-flow-type` were failing because the `no-undef` lint error wasn't being thrown anymore. I removed the tests.

So these changes were required to get our tests working with new parser and eslint. However, I think that with a pull request to `babel-eslint` we will be able to get rid of `define-flow-type` and `use-flow-type` rules entirely. I filed the [issue](https://github.com/babel/babel-eslint/issues/625) few months ago without a response so I will try to create a pull request with needed changes. If they get approved, then we can remove mentioned rules and publish new major version of the plugin. I will update this pull request accordingly.

Btw, there is already `v10.0.0` version of `babel-eslint` but I couldn't use it because [this](https://github.com/babel/babel-eslint/pull/584) commit started to rename `TypeAlias` nodes into `FunctionDeclaration` and it of course broke a lot of our rules.

Because of `eslint-config-canonical` update, I had to apply lint fixes to a lot of files. It kind of messes this pull request, however, I wasn't able to separate lint and bump changes into two commits. the most important files of this pull request are source and test files for `define-flow-type` and `use-flow-type`.


pnevyk:
@gajus This PR was not supposed to be merged yet, hence I prefixed it with WIP. I intended it to be a major change removing the mentioned rules and adding support for newer `babel-eslint` and `eslint`. Nevertheless, I will continue on getting rid of `define-flow-type` and `use-flow-type` as mentioned in #336 in a separate PR.

Issue #361, opened 2018-10-09
Add a rule to disallow `mixed` types


kangax:
There are cases in our app where developers use `mixed` instead of a more exact value. While there are definitely valid scenarios for `mixed` — e.g. on a utility/library level — they're almost never needed on an application level. 

It would be nice to have the rule to disallow them.

This could be another value in a `no-weak-types` or a separate rule.

Hypnosphi:
`mixed` is actually stricter than any other type, as it allows no assumptions about the value. I don't get how can it be considered "weak".

The usecase for app is any function that accepts a callback and doesn't use its return value:

```
function callCb(cb: () => mixed) {
  cb()
}
```

gajus:
https://flow.org/en/docs/types/mixed/

Can be any primitive data type (number, string).

I don't understand the example that you've shared.

Shouldn't it be `void` instead?

```js
function callCb(cb: () => void) {
  cb()
}

```

Hypnosphi:
> Can be any primitive data type

It doesn't make the type weak. You can't mistakenly type number as string when using `mixed` (as contrary to `any`). Coercing `mixed` to any other type always throws an error.

> Shouldn't it be void instead?

then it will error on a callback that returns something (a promise, for example)

The example is from docs: https://flow.org/en/docs/types/functions/#toc-function-type

Hypnosphi:
> Does it make sense to have this logic as a separate rule?

Sure, people may want to only accept callbacks with a specific return value in their app.

leoselig:
I would sincerely propose to revert this change. As @Hypnosphi pointed out ciorrectly, `mixed` is not in any way weak. It is **the** proper way to demand code dealing with the value to deal with **all possible values**. 
This is the absolute opposite to:
- `Function` (call it with what you want)
- `Object` (it's an object, but `you.can.access.anything.you.want`)
- `any` (essentially: skip type-checking anything you do with it until casting it to a concrete type again)

gajus:
> I would sincerely propose to revert this change. 

Agreed. Done.

kangax:
@leoselig @gajus I find this to be a matter of semantics. In Flow lingo, `Function`, `Object`, and `any` are called "unsafe". There's no such thing as "weak", as far as I know, so it really is up to the rule to determine what to consider "weak" and what not to. Alternatively, the rule could be called "no-unsafe-types" and then we wouldn't need to argue about any of this :)

[...]

PR #362, opened 2018-10-12
Add mixed to no-weak-types

kangax:
Fixes https://github.com/gajus/eslint-plugin-flowtype/issues/361

### DIFF
--- a/README.md
+++ b/README.md
@@ -1568,19 +1568,20 @@ The following patterns are not considered problems:
 <a name="eslint-plugin-flowtype-rules-no-weak-types"></a>
 ### <code>no-weak-types</code>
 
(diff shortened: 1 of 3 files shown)


kangax:
@gajus my concern is that this is not backward-compatible (from what I can tell) since `mixed` will now be reported unless disabled explicitly.

gajus:
Sorry, I meant as a configuration.

kangax:
@gajus you mean as a separate rule?

kangax:
Hm, not sure I understand. Isn't the rule already accepting a configuration? I just added another option for mixed there.

gajus:
> Hm, not sure I understand. Isn't the rule already accepting a configuration? I just added another option for mixed there.

Sorry. Ignore my comments. :-)

PR #366, opened 2018-10-26
docs: Add docs about require-compound-type-alias to readme

pnevyk:
I apologize but I didn't add the docs of the rule to the readme. I should have tried to generate the docs locally...


gajus:
Thank you. I have added a git hook to auto generate the docs.

d070ac32703c2c3337a84e554508966a40b5527b

pnevyk:
@gajus Maybe a custom lint script might be valuable for checking if for every (new) rule there is a test file and doc file and they are also correctly included in corresponding index files (`tests/rules/index.js` for tests, `.README/README.md` for docs). This could be run during pre-commit hook along with other lints.

If you think it would be useful, I can provide an implementation.

gajus:
ESLint rule sounds excessive... perhaps a simple ad hoc script?

PR #368, opened 2018-10-27
Tests docs check script

pnevyk:
This PR adds a script that checks if there are tests and docs for each rule and they are included in their corresponding "index" files. It was discussed in #366. What exactly is checked is documented in the header of the script. I added this check to pre-commit hook.

Thanks to this script I discovered that three rules have no assertions in readme and that for one rule tests were not executed. I fixed these issues.

Side note: The new post-commit hook crashes when the current branch is not in `.git/config` yet, which I think is before the push to a remote server. This crash comes from gitdown and transitively from gitinfo. Also in [contributing guide](https://github.com/gajus/eslint-plugin-flowtype/blob/200868729916286e711347877fbef76b25d7b5e0/CONTRIBUTING.md) there is a sentence that the documentation is built using CI but it seems that it is not true anymore (imo it was removed in [this](https://github.com/gajus/eslint-plugin-flowtype/commit/e9395961157a7391d89b9eeab9c5d0b6d4d3de9f#diff-354f30a63fb0907d4ad57269548329e3) commit). I think that generating docs in CI is better approach than in post commit hook but that's just my hint.

### DIFF
--- a/.README/rules/no-flow-fix-me-comments.md
+++ b/.README/rules/no-flow-fix-me-comments.md
@@ -19,4 +19,4 @@ This rule takes an optional RegExp that comments a text RegExp that makes the su
 }
 ```
 
-<!-- assertions no-flow-fix-me-comments -->
+<!-- assertions noFlowFixMeComments -->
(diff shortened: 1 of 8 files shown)


gajus:
Looks great. Thank you.

Is it ready to be merged?

pnevyk:
If you like it then I think that yes. Maybe the script can be added to travis script but that's up to you. And if you can in some time resolve the side note I wrote in the description of this PR, it would be great.

PR #379, opened 2018-11-30
Update README.md

iliyaZelenko:
I think it is more convenient than waiting for the installation of each package.


iliyaZelenko:
It's strange why the whole file has changed, I changed only two lines:

> ```
> # Or all at once: 
> npm install eslint babel-eslint eslint-plugin-flowtype --save-dev
> ```

iliyaZelenko:
Sorry for this, did not know that this is possible. [Created another PR](https://github.com/gajus/eslint-plugin-flowtype/pull/380).

PR #387, opened 2019-02-15
fix(requireReturnType): fix async generator bug [ci failure is irrelevant]

### LINKED ISSUE #386: require-return-type says "must annotate undefined return type" on an async generator function
```js
const resolvers = {
  TagState: {
    async *subscribe(): AsyncIterable<TagState> {
      await ...
      yield {...}
    }
  }
}
```

```
error  Must annotate undefined return type  flowtype/require-return-type
```

Annotating `: void` of course makes flow complain about the return type.

```
      2,
      "always",
      {
        "annotateUndefined": "always",
        "excludeArrowFunctions": true,
        "excludeMatching":[
          "constructor", "componentDidMount",
          "componentDidUpdate", "componentWillReceiveProps",
          "componentWillUnmount"
        ]
      }
    ],
```

eslint: 5.13.0
eslint-plugin-flowtype: 3.4.1

jedwards1211:
fix #386

There were no test cases for even sync generators with `require-return-type`, so I added
those too.


jedwards1211:
@gajus CI only failed because yarn refuses to install on node 6 because of `engines.node` in the `package.json`.  I made a PR #388 to fix this

PR #391, opened 2019-03-26
feat: add rule to make sure that object type that is spread has exact type

Kiwka:
This rule will add possibility to enforce spreading only exact object types, that are explicitly exact. 

When object type is not exact and is spread inside other object annotation, it can lead to resulted type to be dependent on the place, where object type was spread. For example, this case, that from first look is invalid https://flow.org/try/#0C4TwDgpgBAygFgQwE4QCYAUkHtJOASwgGcoBeKAbwCgooAbCAOwHNg4AuKRgVwFsAjCEgA0NLgl4RORYEnwtRAXwDcVKqEhQAchADumHEILEylMQmZSufQSLEA6R-GRoDuY0SWqqAYyyMZKFwif04dfWx3QhJyagBICysAciw6VCSlIA
actually doesn't cause any flow errors. 

One of the way to prevent such cases, is to spread only exact object types.


Kiwka:
👋 @gajus. What can I do to make this new feature to be added?

gajus:
Thank you

PR #392, opened 2019-04-02
feat: support interfaces for delimiter-dangle

### LINKED ISSUE #381: delimiter-dangle triggers on interface
In an `interface` I don't think this should trigger the `delimiter-dangle` rule. Mostly because `prettier` keeps adding in the semi-colon and in this situation it is more like a class definition than an object definition so I think the semi-colon on `renderNode` should be there.

```js
interface RenderMarkInterface {
  renderMark: (props: *) => Element<*> | Node | null;
  renderNode?: null; // This line throws "Unexpected trailing delimiter (flowtype/delimiter-dangle)
}
```

SimeonC:
fix #381


gajus:
Thank you.

PR #399, opened 2019-05-02
feat(require-return-type): add annotateUndefined: 'ignore' option

### LINKED ISSUE #398: feature request: annotateUndefined: 'ignore'
I've got more important things to worry about than whether undefined return types are annotated or not, I just want to turn off all errors about whether or not undefined return types are annotated, so I can work around the bug with async functions returning `Promise<void>` and get on with my day...

jedwards1211:
fix #398


jedwards1211:
Thanks!  It's great to see you're using semantic-release, I'm surprised how few people use it

PR #402, opened 2019-05-06
Don't crash when encountering unparameterized Arrays

zpao:
While these are disallowed in Flow, `typeof Array` is a valid type,
which could trigger this.

Rather than try to enumerate other possible cases, I simply check to
make sure we don't crash when we hit them.

### DIFF
--- a/tests/rules/assertions/arrayStyleSimpleType.js
+++ b/tests/rules/assertions/arrayStyleSimpleType.js
@@ -128,6 +128,16 @@ export default {
           onlyFilesWithFlowAnnotation: true
         }
       }
+    },
+
+    // While this isn't valid flow, we shouldn't disallow it.
+    {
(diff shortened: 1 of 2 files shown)


tests/rules/assertions/arrayStyleSimpleType.js
```diff
@@ -128,6 +128,16 @@ export default {
           onlyFilesWithFlowAnnotation: true
         }
       }
+    },
+
+    // While this isn't valid flow, we shouldn't disallow it.
```
gajus: What is the use case for this if it isn't valid Flow?

zpao: Mostly just a sanity check to check that the code won't break. I'm not attached to it at all.

PR #403, opened 2019-05-07
 fix: array-simple-type should fail gracefully when Array is missing a type parameter

bcherny:
This was causing my linter to crash.


gajus:
Duplicate of https://github.com/gajus/eslint-plugin-flowtype/pull/402.

bcherny:
Beat me to it by two hours :)

PR #406, opened 2019-05-23
Fix requireReadOnlyReactProps

kangax:
- `$FlowFixMe` is ignored
- `{||}` is considered to be read-only

### DIFF
--- a/src/rules/requireReadonlyReactProps.js
+++ b/src/rules/requireReadonlyReactProps.js
@@ -100,7 +100,6 @@ const create = (context) => {
 
       if (currentNode.params[0].type === 'Identifier' &&
           (typeAnnotation = currentNode.params[0].typeAnnotation)) {
-
         if ((identifier = typeAnnotation.typeAnnotation.id) &&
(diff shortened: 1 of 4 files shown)


src/rules/requireReadonlyReactProps.js
```diff
@@ -1,6 +1,7 @@
 const schema = [];
 
 const reComponentName = /^(Pure)?Component$/;
+const reReadOnly = /^\$(ReadOnly|FlowFixMe)$/;
```
mrtnzlml: `$FlowFixMe` was just an example. It can be anything defined in `suppress_type`.

kangax: how would we know what's defined there (from within a plugin rule)?

mrtnzlml: You don't. This is why I was suggesting to not throw an error in cases where you don't know it's wrong for sure. :)

pnevyk: Maybe allow to specify the suppress type name in an option of the rule?

PR #408, opened 2019-06-10
feat: add require-inexact-type rule

goodmind:
Duplicate of same lint in new Flow version
This syntax exists from Flow v0.84


mrtnzlml:
Hi @goodmind! I tried this new lint rule on our monorepo and seems like it reports few false positives. For example, have a look at this file:

```js
// @flow strict

export interface ILogger {
  log(...message: $ReadOnlyArray<string>): void;
  warn(...message: $ReadOnlyArray<string>): void;
  error(...message: $ReadOnlyArray<string>): void;
}
```

It reports the following error:

```text
  3:26  error  Type must be explicit inexact  flowtype/require-inexact-type
```

It doesn't sound right (fails for basically every interface). Is it a bug or problem in our code? I see this case is not covered in your tests and I am not sure how should I fix it. Thanks for checking it.

goodmind:
@mrtnzlml fixed

PR #414, opened 2019-06-24
Valid flow strict annotation

xunuoi:
## Support flow strict annotation

- Added eslint rule for `@flow strict` and its fixer
- Ref, https://github.com/gajus/eslint-plugin-flowtype/issues/364

### Examples
```js

// Options: ["always",{"annotationStyle":"line","strict":true}]
// @flow
// Message: Strict Flow file annotation is required, should be `// @flow strict`

```

PR #415, opened 2019-06-24
Fix @noflow support for strict

xunuoi:
## Fix the strict rule for `@noflow`

- When use `@noflow`, it won't check strict flow rule now
- Ref, https://github.com/gajus/eslint-plugin-flowtype/pull/414


xunuoi:
@gajus 

Hey, sorry, there is a bug for @noflow in this PR, I created a fixing path here,

Please help check and review again, sorry about that


Thank you very much!

PR #417, opened 2019-07-18
Implement arrow-parens to support type parameters

### LINKED ISSUE #344: False-positive of arrow-parens with generic arrow function
```js
const t = <A>(a: A): A => {
    const b = a + 1;
    return b;
};
```
I am getting:
> error  Expected parentheses around arrow function argument having a body with curly braces  arrow-parens

which doesn't seem right. Applying fix results in invalid code:

```js
const t = (<)A>(a: A): A => {
```

Workaround is rewrite arrow function to classic function.

lyleunderwood:
This should resolve #344.

I did this quick and dirty, and I'm not sure if it's the correct approach. What I did was:

1. Copied `arrow-parens` rule and tests into `eslint-plugin-flowtype` from `eslint`,
2. Resolved various linting issues and glued code together to work with `eslint-plugin-flowtype` environment,
3. Added tests specific to flow support,
4. Modified behavior of `arrow-parens` rule so that the tests passed.

Let me know if there's a better way to override an eslint rule. Also, I'm not sure if this should go into the recommended config or whatever.

This may also be of interest to https://github.com/typescript-eslint/typescript-eslint/issues/14.

### DIFF
--- a/.README/rules/arrow-parens.md
+++ b/.README/rules/arrow-parens.md
@@ -1,0 +1,18 @@
+### `arrow-parens`
+
+_The `--fix` option on the command line automatically fixes problems reported by this rule._
+
+Enforces the consistent use of parentheses in arrow functions.
+
+This rule has a string option and an object one.
+
+String options are:
+
+- `"always"` (default) requires parens around arguments in all cases.
+- `"as-needed"` enforces no braces where they can be omitted.
+
+Object properties for variants of the `"as-needed"` option:
+
+- `"requireForBlockBody": true` modifies the as-needed rule in order to require parens if the function body is in an instructions block (surrounded by braces).
+
+<!-- assertions arrowParens -->
--- a/src/configs/recommended.json
+++ b/src/configs/recommended.json
@@ -18,9 +18,9 @@
     "flowtype/no-types-missing-file-annotation": 2,
     "flowtype/no-weak-types": 0,
     "flowtype/require-parameter-type": 0,
+    "flowtype/require-readonly-react-props": 0,
     "flowtype/require-return-type": 0,
     "flowtype/require-valid-file-annotation": 0,
-    "flowtype/require-readonly-react-props": 0,
     "flowtype/semi": 0,
     "flowtype/space-after-type-colon": [
       2,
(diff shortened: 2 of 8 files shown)


src/configs/recommended.json
```diff
@@ -18,9 +18,9 @@
     "flowtype/no-types-missing-file-annotation": 2,
     "flowtype/no-weak-types": 0,
     "flowtype/require-parameter-type": 0,
+    "flowtype/require-readonly-react-props": 0,
     "flowtype/require-return-type": 0,
     "flowtype/require-valid-file-annotation": 0,
-    "flowtype/require-readonly-react-props": 0,
```
lyleunderwood: I think a post-commit hook modified this file?

src/rules/arrowParens.js
```diff
@@ -0,0 +1,169 @@
...
+          firstTokenOfParam.range[0],
+          closingParenToken.range[1]
+        ], `${shouldAddSpaceForAsync ? ' ' : ''}${paramToken.value}`);
+      };
+
+      // Type parameters without an opening paren is always a parse error, and
+      // can therefore be safely ignored.
+      if (node.typeParameters) {
+        return;
+      }
```
lyleunderwood: This is the relevant bit. My assumption is that if the arrow function has type parameters, then we can safely ignore it because any arrow function with type parameters and no parens is always a parse error.

.README/rules/arrow-parens.md
```diff
@@ -0,0 +1,18 @@
+### `arrow-parens`
+
+_The `--fix` option on the command line automatically fixes problems reported by this rule._
+
+Enforces the consistent use of parentheses in arrow functions.
+
+This rule has a string option and an object one.
```
lyleunderwood: Grabbed all this stuff from the `eslint` documentation [here](https://eslint.org/docs/rules/arrow-parens).

gajus:
Thank you. LGTM

Mr-Wallet:
@gajus Can you help me get started with running tests with Flow enforcement? I don't see any mention of `flow-bin` in the `package.json` and when I write a test as `const x: number = null;` it passes.

PR #422, opened 2019-07-28
Fix new lint errors now occurring

lukeapage:
I started trying to make changes but lint fails, because https://github.com/gajus/eslint-config-canonical/releases/tag/v17.2.0 causes a whole load of errors on a new npm install.
This PR just runs eslint --fix and commits it.


lukeapage:
@gajus not sure if you noticed but 4 failures were merged after this so the master build currently fails.

PR #426, opened 2019-08-09
Add --fix support to require-indexer-name

goodmind:


goodmind:
Also looks like README doesn't regenerates with commit hooks

gajus:
Disabled the auto generation of the docs as it was causing too much noise in the commit log. Must run `npm run create-readme` before submitting PR.

PR #440, opened 2020-01-10
Add fixer to invalid file annotation style message

TrySound:
This helps to make annotation style consistent gradually when touching files.


joanrodriguez:
👍🏼

gajus:
It appears that this would break `@flow strict`

https://flow.org/en/docs/strict/

TrySound:
@gajus How it will break `@flow strict`? I pass `annotationValue` not hardcoded string. I added test to make it clear.

Issue #376, opened 2018-11-16
require-exact-type only looks at top-level objects


kangax:
with `require-exact-type` enabled:

```js
type Bar = { baz: string }; // reports error
type Foo = {| bar: Bar |};
```

however:

```js
type Foo = {| bar: { baz: string } |}; // no error
```

PR #441, opened 2020-03-19
feat: require-exact-type detects nested objects

Marcosld:
require-exact-type should now detect all object type declarations, not just type aliases. It will also check for nested notations (#376 ).

Also added a fixer.


package.json
```diff
@@ -19,7 +19,7 @@
     "chai": "^4.2.0",
     "eclint": "^2.8.1",
     "eslint": "^5.13.0",
-    "eslint-config-canonical": "^17.3.4",
+    "eslint-config-canonical": "^18.1.1",
```
Marcosld: This dependency was causing an error because in this commit https://github.com/benmosher/eslint-plugin-import/commit/1a3a12869651bec54d567af8dc0e7cb2a801d41f they added a validation for a rule they were using.

src/rules/sortKeys.js
```diff
@@ -88,7 +88,7 @@ const generateOrderedList = (context, sort, properties) => {
       const beforePunctuator = source.getTokenBefore(nextPunctuator, {
```
Marcosld: Reported by rules when I updated the plugin (prefer slice to substring)

tests/rules/assertions/arrowParens.js
```diff
@@ -7,199 +7,199 @@ export default {
     // "always" (by default)
```
Marcosld: Keys were unordered (linting rule)

Marcosld:
@gajus

PR #442, opened 2020-03-24
Fix/require exact type nested

Marcosld:
Just found a bug for #441. Unfortunately AST Node type for Interfaces first child is an ObjectTypeAnnotation too, thus we have to check for that case and ignore it.


Marcosld:
@gajus Sorry for the noise, just found out this bug today :/

PR #451, opened 2020-06-29
Add respect to `@noflow` annotation

### LINKED ISSUE #450: require-return-type should not trigger, if file is annotated with @noflow
`require-return-type` should not trigger, if file is annotated with `@noflow`.
This can be workarounded by settings: `{"flowtype":{"onlyFilesWithFlowAnnotation":true}}`, but in such case `require-valid-file-annotation` stops working.

andrejs-sisojevs-accenture:
Fixes #450 
Added respect to `@noflow` annotation.
Notice: `npm run create-readme` didn't work for me - failed with error `Branch ("master") definition does not exist in the config`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -88,7 +88,7 @@ export default {
     recommended,
   },
   rules: _.mapValues(rules, (rule, key) => {
-    if (key === 'no-types-missing-file-annotation') {
+    if (['no-types-missing-file-annotation', 'require-valid-file-annotation'].includes(key)) {
       return rule;
     }
 
--- a/src/utilities/isNoFlowFileAnnotation.js
+++ b/src/utilities/isNoFlowFileAnnotation.js
@@ -1,0 +1,17 @@
+import _ from 'lodash';
+
+const FLOW_MATCHER = /^@noflow$/;
+
+export default (comment, strict) => {
+  // The flow parser splits comments with the following regex to look for the @flow flag.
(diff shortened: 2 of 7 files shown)


src/index.js
```diff
@@ -88,7 +88,7 @@ export default {
     recommended,
   },
   rules: _.mapValues(rules, (rule, key) => {
-    if (key === 'no-types-missing-file-annotation') {
+    if (['no-types-missing-file-annotation', 'require-valid-file-annotation'].includes(key)) {
```
andrejs-sisojevs-accenture, line 91: Had to change the way how `require-valid-file-annotation` works with `onlyFilesWithFlowAnnotation` setting, because of breaking change in `checkFlowFileAnnotation.js`

src/utilities/isNoFlowFileAnnotation.js
```diff
@@ -0,0 +1,17 @@
...
+  // The flow parser splits comments with the following regex to look for the @flow flag.
+  // See https://github.com/facebook/flow/blob/a96249b93541f2f7bfebd8d62085bf7a75de02f2/src/parsing/docblock.ml#L39
+  return _.some(comment.split(/[\t\n\r */\\]+/), (commentPart) => {
+    const match = commentPart.match(FLOW_MATCHER);
+
+    if (match === null) {
+      return false;
+    }
+
+    return !strict || match[0] === '@noflow';
```
andrejs-sisojevs-accenture, line 15: This file is a copy of "isFlowFileAnnotation.js" with only changes in regex and on this line

andrejs-sisojevs-accenture:
@gajus is there any way I can help advance this PR?

PR #454, opened 2020-08-21
fix: declare parsers as optional peer dependencies

ylemkimon:
The [test](https://github.com/ylemkimon/eslint-plugin-flowtype-babel-eslint/actions/runs/218763178) shows `babel-eslint-plugin` is compatible with the latest of all major version of `babel-eslint`, so `*` is used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,17 @@
   "main": "./dist/index.js",
   "name": "eslint-plugin-flowtype",
   "peerDependencies": {
+    "@babel/eslint-parser": "^7.0.0",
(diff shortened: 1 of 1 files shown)


ylemkimon:
_By @kaicataldo in https://github.com/babel/babel/issues/11989#issuecomment-693892947_
> The hard break was intentional here because we've changed how the parser works and there are significant breaking changes. I'm in favor of making things fail faster in this case so that people are aware of the changes if/when they upgrade.

I think declaring both of them as optional peer dependencies is the best way to go.

Issue #467, opened 2021-03-11
Rule 'require-readonly-react-props' not compatible with implicit exact objects


mrtnzlml:
Hi! 👋 We are using `exact_by_default=true` for a while now and [this commit](https://github.com/facebook/relay/commit/6fa0b0d83b45af67d086e5b850b4b49915eccb4a) motivated me to investigate what it takes to migrate from explicit exact objects `{| |}` to implicit exact objects `{ }`. It's fairly straightforward and it mostly works well.

However, we are using `flowtype/require-readonly-react-props` rule which doesn't seem to be compatible with implicit exact objects. It understands only `{||}` syntax and complains about empty props not being read-only, see: https://github.com/gajus/eslint-plugin-flowtype/blob/e1d5d049ae12a024f6007772a7d57acef254ac3d/src/rules/requireReadonlyReactProps.js#L29-L32

I wanted to discuss if you'd be open to changing it and if so, how should the updated rule look like? 🤔 Maybe something like this?

```json5
{
    "rules": {
        "flowtype/require-readonly-react-props": ["error", {
            "implicitExactTypes": true // ???
        }]
    }
}
```

Thanks! 🙂

Brianzchen:
Since there have been no replies, I'll just chime in that the change to the rule looks fine to me. I personally don't use it cause it's a bit extreme and it's not enabled by default from the recommended config so it would never impact me, but given that you can't even use it with an existing rule in this plugin with this one, it makes sense that it gets fixed somehow.

mrtnzlml:
Here is my attempt to fix this issue: https://github.com/gajus/eslint-plugin-flowtype/pull/471

PR #470, opened 2021-03-15
docs: regenerate README.md

mrtnzlml:
Followup after https://github.com/gajus/eslint-plugin-flowtype/commit/fd838c3dce0e5917d7a26ed8074645b2d510cd68

I didn't generate the docs because I misunderstood the contribution guide. This commit fixes it and updates the contribution guide a little bit.

PR #471, opened 2021-03-15
feat: add support for implicit exact Flow types

mrtnzlml:
This commit adds a support for implicit exact Flow types. This is especially useful when using `exact_by_default=true` Flow option and migrating towards `{ }` and `{ ... }` syntax (Flow standard is now to use `{ }` for strict objects and `{ key: value, ... }` for open objects).

Closes: https://github.com/gajus/eslint-plugin-flowtype/issues/467

TODO:

- [x] rebase after https://github.com/gajus/eslint-plugin-flowtype/pull/470
- [x] regenerate docs

### DIFF
--- a/src/rules/requireReadonlyReactProps.js
+++ b/src/rules/requireReadonlyReactProps.js
@@ -1,4 +1,16 @@
-const schema = [];
+import _ from 'lodash';
+
+const schema = [
+  {
+    additionalProperties: false,
+    properties: {
+      useImplicitExactTypes: {
+        type: 'boolean',
+      },
+    },
+    type: 'object',
+  },
+];
 
 const reComponentName = /^(Pure)?Component$/;
 const reReadOnly = /^\$(ReadOnly|FlowFixMe)$/;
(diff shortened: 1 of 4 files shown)


tests/rules/assertions/requireReadonlyReactProps.js
```diff
@@ -194,5 +210,13 @@ export default {
...
       code: 'function Foo(props: {||}) { return <p /> }',
     },
+    {
+      code: 'function Foo(props: {}) { return <p /> }',
+      options: [
+        {
+          implicitExactTypes: true,
+        },
+      ],
+    },
```
Brianzchen, line 220: I'm just thinking for an extra valid case, would you wanna use `{| |}` syntax and with `implicitExactTypes: true,`. Just to test that it doesn't blow up for an edge case?

src/rules/requireReadonlyReactProps.js
```diff
@@ -21,13 +33,16 @@ const isReactComponent = (node) => {
...
 
-  // we consider `{||}` to be ReadOnly since it's exact AND has no props
-  if (node.exact && node.properties.length === 0) {
+  // we consider `{}` to be ReadOnly since it's exact AND has no props (when `implicitExactTypes=true`)
+  // we consider `{||}` to be ReadOnly since it's exact AND has no props (when `implicitExactTypes=false`)
+  if (implicitExactTypes === true && node.exact === false && node.properties.length === 0) {
+    return true;
+  } else if (node.exact === true && node.properties.length === 0) {
     return true;
   }
```
Brianzchen, line 47: I'm wondering if this is nicer and more readable, the close duplication in syntax of your two conditions took me a while to notice that `node.exact` `===` `true/false` in both conditions
```suggestion
  if (node.properties.length === 0) {
    if (node.exact === false && implicitExactTypes === true) {
      return true;
    }
    if (node.exact === true) {
      return true;
    }
  }
```

tests/rules/assertions/requireReadonlyReactProps.js
mrtnzlml: Sure, good idea. 👍

Brianzchen (approved):
LGTM

Brianzchen:
@gajus

Brianzchen:
@gajus bump

mrtnzlml:
Hi, is there anything I can improve in this PR to make it mergeable?

I would also like to offer my help with the maintenance of this project if needed. We are using this plugin for years and it's always been a pleasure. I would not like to see it slowly die. 😎

gajus:
I cannot recall the reason I did not merge this the first few times it was bumped. Sorry about that.

PR #472, opened 2021-03-15
feat: migrate a new rule 'use-read-only-spread'

mrtnzlml:
This commit adds a new rule 'use-read-only-spread' originally developed and used in [`adeira/universe`](https://github.com/adeira/universe/blob/91ba39d4f59ac0376121b67e443f474a6ade4feb/src/eslint-plugin-adeira/src/rules/flow-use-readonly-spread.js).

It prevents users from accidentally creating an object which is no longer read-only because of Flow spread operator.

TODO:

- [x] rebase after https://github.com/gajus/eslint-plugin-flowtype/pull/470 and renegerate docs


mrtnzlml:
Hi @gajus! Here is another addition to this plugin if you have some extra time for a review. Thanks! 😎

gajus:
Thank you

PR #475, opened 2021-04-08
feat: Added rule to enforce quotes type

### LINKED ISSUE #473: Flowtype quotes rule
There is no rule to enforce single quotes or double quotes for type objects or aliases.

Any of these are valid and doesn't respect eslint's base `quotes` rule.
```
type A = 'test';
type B = "test";
type T = {
  test: "hello" | 'test',
  t: 'hello',
};
```

angelica-bocanegra:
closes #473 
Added new rule `quotes`  that enforces consistent quote types around strings.

Options: `double` (default) or `single`

Double (default):
```
// Correct
type Y = {foo: "bar"};
type T = { test: "hello" | "world" }

// Incorrect
type N = { foo: 'bar' };
type T = { test: "hello" | 'world' }

```

Single:
```
// Correct
type Y = {foo: 'bar'};
type T = { test: 'hello' | 'world' }

// Incorrect
type N = { foo: 'bar' };
type T = { test: "hello" | 'world' }

```
@gajus

### DIFF
--- a/README.md
+++ b/README.md
@@ -2756,44 +2756,44 @@ type Foo = { a: Foo, b: Bar }
 <a name="eslint-plugin-flowtype-rules-quotes"></a>
 ### <code>quotes</code>
 
-Enforces single quotes or double quotes.
+Enforces single quotes or double quotes around string literals.
 
 <a name="eslint-plugin-flowtype-rules-quotes-options-3"></a>
 #### Options
 
 The rule has string options of:
 
-* `"double"` (default) requires double quotes around strings.
-* `"single"` requires single quotes around strings.
+* `"double"` (default) requires double quotes around string literals.
+* `"single"` requires single quotes around string literals.
 
 The following patterns are considered problems:
 
 ```js
 type T = 'hi'
-// Message: Strings must use double quote.
+// Message: String literals must use double quote.
 
 // Options: ["double"]
 type T = { test: 'hello' | 'test' }
-// Message: Strings must use double quote.
-// Message: Strings must use double quote.
+// Message: String literals must use double quote.
+// Message: String literals must use double quote.
 
 // Options: ["double"]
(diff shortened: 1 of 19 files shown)


src/rules/quotes.js
```diff
@@ -0,0 +1,51 @@
...
+      if (double && sourceCode.text[node.range[0]] !== '"') {
+        // double
+        context.report({
+          fix: (fixer) => {
+            return [
+              fixer.replaceTextRange([node.range[0], node.range[0] + 1], '"'),
+              fixer.replaceTextRange([node.range[1] - 1, node.range[1]], '"'),
+            ];
+          },
+          message: 'Strings must use double quote.',
```
Brianzchen, line 27: I think we should name these as `String literals` not just `Strings`

mrtnzlml:
@angelica-bocanegra It might be a good idea to add this rule here as well: https://github.com/prettier/eslint-config-prettier (https://github.com/prettier/eslint-config-prettier/blob/96efee3dc6d37f9a89d6facb04359a73409cb1e6/index.js#L111-L120) to disable it for Prettier users.

README.md
```diff
@@ -2840,44 +2840,44 @@ type Foo = { a: Foo, b: Bar }
...
+Enforces single quotes or double quotes around string literals.
 
 <a name="eslint-plugin-flowtype-rules-quotes-options-3"></a>
 #### Options
 
 The rule has string options of:
 
-* `"double"` (default) requires double quotes around strings.
-* `"single"` requires single quotes around strings.
+* `"double"` (default) requires double quotes around string literals.
```
Brianzchen, line 2850: Hm I wonder why is default double? Guess same as regular quotes rule?

angelica-bocanegra: Yep keeping it standard

Brianzchen (approved):

PR #477, opened 2021-04-16
fix: support union type in React react-only props rule

mrtnzlml:
This change fixes the following (simplified) example from our application:

```
type Props =
  | {
      +children: React$Node,
      +title: React$Node,
      +withHiddenTitle?: false,
    }
  | {
      +children: React$Node,
      +title?: React$Node,
      +withHiddenTitle: true,
    };

const props: Props = {children:"", withHiddenTitle: true}
props.title = ""; // correct Flow error
```

Flow correctly throws the following error:

```
14: props.title = "";
          ^ Cannot assign empty string to `props.title` because property `title` is not writable. [cannot-write]
```

However, `flowtype/require-readonly-react-props` Eslint rule throws the following (incorrect) error:

```
Props must be $ReadOnly
```

For simplification, all of the objects in the union must be readonly even though it's probably not a strict requirement in Flow. This can be improved later when needed.

### DIFF
--- a/.README/rules/require-readonly-react-props.md
+++ b/.README/rules/require-readonly-react-props.md
@@ -1,6 +1,6 @@
 ### `require-readonly-react-props`
 
-This rule validates that React props are marked as $ReadOnly. React props are immutable and modifying them could lead to unexpected results. Marking prop shapes as $ReadOnly avoids these issues.
+This rule validates that React props are marked as `$ReadOnly`. React props are immutable and modifying them could lead to unexpected results. Marking prop shapes as `$ReadOnly` avoids these issues.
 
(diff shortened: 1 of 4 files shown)


mrtnzlml:
https://flow.org/try/#0C4TwDgpgBACgTgezAZygXgFBSgHygby22IGoBjACwEsAbAEzggDsAuKAJQgEMzgASAHII6EADRFiUEsCrAaENpx78hI8ZOwkA7rIoAJKnRFMAKrPkB+NgDMuNZGIlQAvkTyENUyrQbNF3XkFhR09pcwgrDgCVYPUNbV0DI2YzOQUoYDgAVxDiZwBuDAwyBCZkYCgwRBQ2eCRUNAJvekZWACI20SgdYH1DY1T5Nkyc1yr6gDoZNPQoDsKgA

PR #478, opened 2021-04-16
fix: fix read-only React props error position

mrtnzlml:
Example from our repo before this change:

```
error: Props must be $ReadOnly (flowtype/require-readonly-react-props) at src/kochka.com.mx/src/Layout.js:39:7:
  37 |   return (
  38 |     <>
> 39 |       <Head>
     |       ^
  40 |         <title>KOCHKA café · {props.title}</title>
  41 |       </Head>
  42 |
```

As you can see the error seems to be wrongly positioned. This is even more awkward when you want to suppress this error (and you have to do it on some unrelated line).

After this fix:

```
error: Props must be $ReadOnly (flowtype/require-readonly-react-props) at src/kochka.com.mx/src/Layout.js:28:39:
  26 |     };
  27 |
> 28 | export default function Layout(props: Props): React.Node {
     |                                       ^
  29 |   if (props.withHiddenTitle === true) {
  30 |     invariant(props.title == null, 'Cannot use `title` together with `withHiddenTitle` property.');
  31 |     invariant(
```

### DIFF
--- a/src/rules/requireReadonlyReactProps.js
+++ b/src/rules/requireReadonlyReactProps.js
@@ -144,7 +144,7 @@ const create = (context) => {
 
           context.report({
             message: identifier.name + ' must be $ReadOnly',
-            node,
+            node: identifier,
           });
 
           reportedFunctionalComponents.push(identifier);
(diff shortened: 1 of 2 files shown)


mrtnzlml:
Side note: we are using [`adeira/eslint-fixtures-tester`](https://github.com/adeira/eslint-fixtures-tester) for better Eslint rules testing. It has the line/column testing integrated so it's easy to capture such things. I think it's worth considering it even though it would require more changes to this repo (for example the README generation).

PR #479, opened 2021-04-20
chore: update recommended config to @babel/eslint-parser

### LINKED ISSUE #474: 'babel-eslint' deprecation and rename
Babel's eslint parser has been moved into their main monorepo, and republished under a different name:

* Deprecation on npm: https://www.npmjs.com/package/babel-eslint
* Old repository archived: https://github.com/babel/babel-eslint
* New location: https://github.com/babel/babel/tree/main/eslint/babel-eslint-parser

Because the name is different, eslintrc configuration settings will need to change from 
```
"parser": "babel-eslint",
```
to 
```
"parser": "@babel/eslint-parser",
```
Making this change to this plugin's recommended config seems to me like a breaking change.

`babel-eslint` still seems to work fine, even with recent babels, but it's been a year since it's had updates and I assume something will break eventually.

Brianzchen:
Closes #474

Recommended config now uses @babel/eslint-parser. This should be a breaking change for consumers, the only difference should be that eslint how uses a projects babel config to read files which may cause issues if the babel config has exclusions setup based on environment variables that don't play nice with eslint.

This should be a major version bump.

@wilt-rally @gajus

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "main": "./dist/index.js",
   "name": "eslint-plugin-flowtype",
   "peerDependencies": {
+    "@babel/eslint-parser": "^7.12.0",
     "eslint": "^7.1.0"
   },
   "repository": {
--- a/src/configs/recommended.json
+++ b/src/configs/recommended.json
@@ -1,5 +1,5 @@
 {
-  "parser": "babel-eslint",
+  "parser": "@babel/eslint-parser",
   "plugins": [
     "flowtype"
   ],


Brianzchen:
@gajus thoughts on this?

gajus:
* What's the reason `babel-eslint` is kept as a dependency?
* What's the reason `@babel/eslint-parser` is not a dependency?

Brianzchen:
> * What's the reason `babel-eslint` is kept as a dependency?

`babel-eslint` is a dev-dependency of this project is because the eslint rules of this project don't use itself, the dependency here also isn't what is shipped as a requirement to the user.

> * What's the reason `@babel/eslint-parser` is not a dependency?

This project doesn't use `@babel/eslint-parser` again because of the eslint rules it extends from are not itself. The current version of canonical eslint still uses `babel-eslint` https://github.com/gajus/eslint-config-canonical/blob/v20.1.0/configurations/eslintrc.json#L7. We can upgrade this as part of a separate PR because updating the dev dependencies here is not really the goal of this PR

---

Overall what i'm trying to do here is just update what is used by the consumer so that they get `@babel/eslint-parser` out of the box without needing to override `babel-eslint` not what this project (eslint-plugin-flowtype) internally uses.

Brianzchen:
Fix published to [eslint-plugin-ft-flow](https://www.npmjs.com/package/eslint-plugin-ft-flow)

PR #484, opened 2021-06-27
feat: add new rule enforce-line-break

### LINKED ISSUE #476: Enforce breakline on type definitions
I would like to a rule to enforce breaklines between type definitions or any other statements for that matter.

The rule to start doesn't need any options and should enforce one line at least between type defs when on, off would do nothing.

Examples:
```
// Error
type A = string;
type B = number;

// Fixes to 
type A = string;

type B = number;
```

```
// Error
import * as React from 'react';
type A = string;
const B = () => {};

// Fixes to
import * as React from 'react';

type A = string;

const B = () => {};

// This is also valid
import * as React from 'react';


type A = string;


const B = () => {};
```

angelica-bocanegra:
New rule to enforce line breaks before and after type declarations.

Closes #476 
```
// Error
type A = string;
type B = number;
// Some Comment
type rez = string;

// Fixes to 
type A = string;

type B = number;

// Some Comment
type rez = string;

```

@gajus

### DIFF
--- a/src/rules/enforceLineBreak.js
+++ b/src/rules/enforceLineBreak.js
@@ -1,0 +1,61 @@
+const schema = [];
+
+const create = (context) => {
+  return {
+    TypeAlias (node) {
+      const sourceCode = context.getSourceCode();
+
+      // if type alias has spacing above and below cancel out of checks
+      // Check for comments above the line (if there is a comment then add the space above that instead)
+      if (node.loc.start.line !== 1) {
+        // Check if there are comments above the line
+        if (node.leadingComments) {
+          const lineAboveComment = sourceCode.lines[node.leadingComments[0].loc.start.line - 2];
+          if (lineAboveComment !== '') {
+            context.report({
+              fix (fixer) {
+                return fixer.insertTextBeforeRange(node.leadingComments[0].range, '\n');
+              },
+              message: 'Please enter a line above type declaration',
+              node,
+            });
+          }
+        } else if (!node.leadingComments) {
+          const lineAbove = sourceCode.lines[node.loc.start.line - 2];
+
+          // there are no comments add space above line
+          if (lineAbove !== '') {
+            context.report({
+              fix (fixer) {
+                return fixer.insertTextBefore(node, '\n');
+              },
+              message: 'Please enter a line above type declaration',
+              node,
+            });
+          }
+        }
+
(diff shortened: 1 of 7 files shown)


src/rules/enforceLineBreak.js
```diff
@@ -0,0 +1,59 @@
...
+
+      if (node.loc.start.line !== 1) {
+        if (node.leadingComments && node.leadingComments[0].loc.start.line !== 1) {
+          const lineAboveComment = sourceCode.lines[node.leadingComments[0].loc.start.line - 2];
+          if (lineAboveComment !== '') {
+            context.report({
+              fix (fixer) {
+                return fixer.insertTextBeforeRange(node.leadingComments[0].range, '\n');
+              },
+              message: 'Please enter a line above type declaration',
```
Brianzchen, line 19: I think the error messages general have the tone of `New line required above type declaration`. Should we change to like that?

src/rules/enforceLineBreak.js
```diff
@@ -0,0 +1,59 @@
...
+            });
+          }
+        } else if (!node.leadingComments) {
+          const isLineAbove = sourceCode.lines[node.loc.start.line - 2];
+          if (isLineAbove !== '') {
+            context.report({
+              fix (fixer) {
+                return fixer.insertTextBefore(node, '\n');
+              },
+              message: 'Please enter a line above type declaration',
```
Brianzchen, line 30: Should the message itself be a func that can be reused for both above and below?
```
const breakLineMessage = (direction) => `Please enter a line ${direction} type declaration`;
```

src/rules/enforceLineBreak.js
```diff
@@ -0,0 +1,59 @@
...
+
+const create = (context) => {
+  return {
+    TypeAlias (node) {
+      const sourceCode = context.getSourceCode();
+      if (sourceCode.lines.length === 1) {
+        return;
+      }
+
+      if (node.loc.start.line !== 1) {
```
Brianzchen, line 11: Just curious what does this do?

angelica-bocanegra: This is so you won't get the error 'New line required above type declaration` if the type declaration happens to be on the first line.

Brianzchen (approved):
LGTM

Issue #485, opened 2021-07-22
generic-spacing doesn't support CRLF line breaks


pireads:
The generic-spacing rule doesn't support CRLF line breaks.

```js
// Passes
type X =  Promise<\n    (foo),\n    bar,\n    (((baz)))\n>

// Doesn't pass
type X =  Promise<\r\n    (foo),\r\n    bar,\r\n    (((baz)))\r\n>
```

pireads:
Thank you for getting this in so quickly!

PR #486, opened 2021-07-22
fix: support CRLF line breaks in generic-spacing (#485)

pireads:
The generic-spacing rule doesn't support CRLF line breaks.

```js
// Passes
type X =  Promise<\n    (foo),\n    bar,\n    (((baz)))\n>

// Doesn't pass
type X =  Promise<\r\n    (foo),\r\n    bar,\r\n    (((baz)))\r\n>
```


axosoft-avi:
Closes #485

PR #488, opened 2021-07-25
enforceLineBreak now handles `export type` correctly

### LINKED ISSUE #487: enforce-line-break: issues with auto fixer and `export type`
```
const a = 5;
export type A = string
```

Turns into
```
const a = 5;
export

type A = string;
```

cc: @angelica-bocanegra

Brianzchen:
Closes #487

### DIFF
--- a/src/rules/enforceLineBreak.js
+++ b/src/rules/enforceLineBreak.js
@@ -12,25 +12,28 @@ const create = (context) => {
         return;
       }
 
+      const exportedType = node.parent.type === 'ExportNamedDeclaration';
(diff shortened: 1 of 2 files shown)


Brianzchen:
cc: @gajus @angelica-bocanegra

Brianzchen:
Found that `node.leadingComments` won't work when published, for some reason the test suite still populates the values. And eslint docs recommend `sourceCode.getCommentsBefore()` instead which still lead to reliable tests and now also replicates the real world

https://eslint.org/docs/user-guide/migrating-to-4.0.0#ast-nodes-no-longer-have-comment-properties

angelica-bocanegra (approved):
LGTM

PR #489, opened 2021-07-28
feat: new `enforceSuppressionCode` rule

### LINKED ISSUE #453: Add rule to enforce error codes on flow suppressions
Flow version 0.121.0 allows for more specific error suppressions:
https://medium.com/flow-type/making-flow-error-suppressions-more-specific-280aa4e3c95c

It would be wonderful if we could have a rule which requires these specific error suppressions:

* `// $FlowFixMe` would fail since it has no valid error code specified
* `// $FlowFixMe[incompatible-type]` would pass since it does have a valid error code

Brianzchen:
Closes #453

cc: @christianvuerings @gajus @angelica-bocanegra

### DIFF
--- a/src/rules/enforceSuppressionCode.js
+++ b/src/rules/enforceSuppressionCode.js
@@ -1,0 +1,54 @@
+const schema = [
+  {
+    type: 'string',
+  },
+];
+
+const message = (suppression = '$FlowFixMe') => {
+  return `${suppression} is missing a suppression code`;
+};
+
+const create = (context) => {
+  const isMissingSuppressionCode = function (value) {
+    if (value.startsWith('$FlowFixMe') &&
+        !value.startsWith('$FlowFixMe[') &&
+        !value.endsWith(']')) {
+      return '$FlowFixMe';
+    }
+    if (value.startsWith('$FlowExpectedError') &&
+        !value.startsWith('$FlowExpectedError[') &&
+        !value.endsWith(']')) {
+      return '$FlowExpectedError';
+    }
+
+    return undefined;
+  };
+
+  const handleComment = function (comment) {
+    const value = comment.value.trim().split(' ').filter((arg) => {
+      return arg.length > 0;
+    })[0];
+    const failedType = isMissingSuppressionCode(value);
+
+    if (failedType) {
+      context.report(comment, message(failedType));
+    }
+  };
+
+  return {
+    Program () {
+      context
+        .getSourceCode()
+        .getAllComments()
+        .filter((comment) => {
+          return comment.type === 'Block' || comment.type === 'Line';
+        })
+        .forEach(handleComment);
+    },
+  };
+};
+
+export default {
+  create,
+  schema,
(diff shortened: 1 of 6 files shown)


src/rules/enforceSuppressionCode.js
```diff
@@ -0,0 +1,54 @@
...
+  },
+];
+
+const message = (suppression = '') => {
+  return `${suppression} is missing a suppression code`;
+};
+
+const create = (context) => {
+  const isMissingSuppressionCode = function (value) {
+    const suppressionTypes = ['$FlowFixMe', '$FlowExpectedError'];
```
mrtnzlml, line 13: What about adding there all the newly supported suppressions? See: https://flow.org/en/docs/config/options/#toc-suppress-comment-regex

> Note: In version v0.127.0, the option to specify the suppression comment syntax was removed. $FlowFixMe, $FlowIssue, $FlowExpectedError, and $FlowIgnore became the only standard suppressions.

tests/rules/assertions/enforceSuppressionCode.js
```diff
@@ -0,0 +1,31 @@
+export default {
```
mrtnzlml, line 1: It would be nice to have here some examples of block comments as well, see: https://flow.org/en/docs/errors/

Brianzchen: Cool found issue with block comments so fixed it

Brianzchen:
@mrtnzlml any more thoughts on the change?

tests/rules/assertions/enforceSuppressionCode.js
```diff
@@ -0,0 +1,90 @@
...
+  invalid: [
+    {
+      code: '// $FlowFixMe I am doing something evil here\nconst text = \'HELLO\';',
+      errors: [
+        {
+          message: '$FlowFixMe is missing a suppression code',
```
mrtnzlml, line 7: Flow is warning with the following message when the suppression code is missing:

```text
Suppression is missing a code. Please update this suppression to use an error code: $FlowFixMe[incompatible-cast]
```

I understand that this would be complicated to do in Eslint, but maybe it would be nice to explain how to fix this error? For example:

```text
$FlowFixMe is missing a suppression error code. Please update this suppression to use an error code: $FlowFixMe[…]
```

What do you think? Just a nit.

mrtnzlml (approved):

tests/rules/assertions/enforceSuppressionCode.js
Brianzchen: Yeah no worries, let me look into it later

Brianzchen: Done, agree it's heaps better

mrtnzlml (approved):

Brianzchen:
@gajus bump

Brianzchen:
@gajus Can this be merged?

PR #490, opened 2021-07-31
Object curly newline

### LINKED ISSUE #468: Rule for object curly brace new line
```
type Obj = {
  test: number };
```
The above should error and instead be fixed to
```
type Obj {
  test: number,
};
```

Brianzchen:
Closes #468

### DIFF
--- a/tests/rules/assertions/objectTypeCurlyNewline.js
+++ b/tests/rules/assertions/objectTypeCurlyNewline.js
@@ -59,13 +59,43 @@ export default {
 
       // output: 'type obj = {"foo": "bar"}',
     },
+
+    // Multiline
+    {
+      code: 'type obj = { "foo": "bar"\n}',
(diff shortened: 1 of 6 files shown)


tests/rules/assertions/objectTypeCurlyNewline.js
```diff
@@ -0,0 +1,93 @@
+export default {
+  invalid: [
+    // Never
+    {
```
angelica-bocanegra, line 4: It would be good to have some tests when there object has multiple properties

Brianzchen: Yes I can, just to note though this rule does not cater to multi-lining multiple properties, that'll be handled by a separate rule hopefully in the near future just like standard objects https://eslint.org/docs/rules/object-property-newline

PR #492, opened 2021-08-02
feat: Interface Id Match rule

### LINKED ISSUE #483: type-id-match doesn't work on interface
https://github.com/gajus/eslint-plugin-flowtype#type-id-match This rule only works with type but not interface, eg

eslintrc.js
```
...
'flowtype/type-id-match': ['error', '^_?([A-Z][A-Za-z0-9]*T|Props)$'],
....
```

```
type MyTypeT = {}; // works
type MyType = {}; // errors
interface MyType {}; // does not error but i would like it to
```

I think it should be a new rule as opposed to being enforced on `type-id-match`. Maybe named `interface-id-match` with the same usage

Jermayy:
Closes #483 

cc: @gajus @angelica-bocanegra @Brianzchen


Brianzchen (approved):
LGTM Thanks!

Issue #502, opened 2021-09-14
Bug: incorrect `no-undef` error for Flow Enums


mrtnzlml:
I know this sounds like it's not a problem with this plugin but bear with me.

Problematic code:

```js
// @flow strict

enum Status {
  Active,
  Paused,
  Off,
}
```

This code uses the new [Flow Enums](https://flow.org/en/docs/enums/). Eslint throws the following errors:

```text
  3:6  error  'Status' is not defined  no-undef
  4:3  error  'Active' is not defined  no-undef
  5:3  error  'Paused' is not defined  no-undef
  6:3  error  'Off' is not defined     no-undef
```

It should not throw anything (at least not `no-undef`).

We already have [`define-flow-type`](https://github.com/gajus/eslint-plugin-flowtype/tree/4265b274c7ba4e489a1e8f2dae56a25e2cb03c9e#eslint-plugin-flowtype-rules-define-flow-type) that seems to be solving the issue for other Flow specifics ([code](https://github.com/gajus/eslint-plugin-flowtype/blob/4265b274c7ba4e489a1e8f2dae56a25e2cb03c9e/src/rules/defineFlowType.js)). I _think_ we have to extend this rule to support Flow Enums correctly.

mrtnzlml:
This issue seems to be fixed in the newly released descendant of this plugin: https://github.com/flow-typed/eslint-plugin-ft-flow

PR #512, opened 2021-10-29
feat: add support for enum type

maxclaus:
Add support for the enum type usage https://flow.org/en/docs/enums/.

Fixes https://github.com/gajus/eslint-plugin-flowtype/issues/502.

Something I found during the tests, if you use `@babel/plugin-syntax-flow` plugin, then the `babel-plugin-transform-flow-enums` plugin must be placed before `@babel/plugin-syntax-flow`, otherwise the parser won't recognize the enum type. But, I think you could even remove `@babel/plugin-syntax-flow` if you are using `babel-plugin-transform-flow-enums`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "@babel/eslint-parser": "^7.15.8",
     "@babel/node": "^7.15.8",
     "@babel/plugin-proposal-object-rest-spread": "^7.15.6",
-    "@babel/plugin-syntax-flow": "^7.14.5",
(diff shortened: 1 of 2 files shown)


maxcnunes:
hey @gajus, when you have some time, could you please check this PR? thanks 🙇

maxclaus:
Closing this in favor of https://github.com/flow-typed/eslint-plugin-ft-flow which has implemented it already

PR #515, opened 2021-11-03
New rule: `no-flow-fix-me-in-strict-files`

nvie:
This rule validates that no error suppression comments (e.g. `$FlowFixMe`) are used in `// @flow strict` (or `// @flow strict-local`) files.

This codifies the best practices [as documented here](https://flow.org/en/docs/strict/#toc-adoption):

Screen Shot 2021-11-03 at 09 10 04@2x

It's a simple rule without any options.

---

@gajus I followed all of the contributing guidelines, adding documentation and tests, but I failed at the last step while running `npm run create-readme`. See [my error log](https://gist.github.com/nvie/113e8d4053beb080a5d29d9b389598c3) here.

### DIFF
--- a/src/rules/noFlowFixMeInStrictFiles.js
+++ b/src/rules/noFlowFixMeInStrictFiles.js
@@ -1,0 +1,43 @@
+const FLOW_STRICT_MATCHER = /^\s*@(?:no)?flow\s*strict(?:-local)?\s*$/u;
+
+const isStrictFlowFile = (context) => {
+  return context
+    .getAllComments()
+    .some((comment) => {
+      return FLOW_STRICT_MATCHER.test(comment.value);
+    });
+};
+
+const message =
+  'No suppression comments are allowed in "strict" Flow files. Either remove the error supression, or lower the strictness of this module.';
+
+const create = (context) => {
+  if (!isStrictFlowFile(context)) {
+    // Skip this file - nothing to check here
+    return {};
+  }
+
+  return {
+    Program: () => {
+      const comments = context
+        .getSourceCode()
+        .getAllComments()
+        .filter((node) => {
+          return node.type === 'Block' || node.type === 'Line';
+        });
+
+      for (const comment of comments) {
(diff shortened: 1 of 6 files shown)


src/rules/noFlowFixMeInStrictFiles.js
```diff
@@ -0,0 +1,43 @@
...
+    Program: () => {
+      const comments = context
+        .getSourceCode()
+        .getAllComments()
+        .filter((node) => {
+          return node.type === 'Block' || node.type === 'Line';
+        });
+
+      for (const comment of comments) {
+        if (/\$FlowFixMe/iu.test(comment.value)) {
```
Brianzchen: just like I created in a different rule, you may want to consider other suppression error names like `FlowExpectedError` which is my preference
ref: https://github.com/gajus/eslint-plugin-flowtype/pull/489/files#diff-ca0fd280a4ec7ba166e4acdb000cdad6a3f6a000d71ad8632113895e58774260R13

nvie: Happy to! Are you suggestion I copy/paste this definition in here too?

```js
const suppressionTypes = ['$FlowFixMe', '$FlowExpectedError', '$FlowIssue', '$FlowIgnore'];
```

Or is there a central "utility file" I could import this definition from?

Brianzchen: I think just copy and paste for now. Since my pr isn't merged either, room for some utility constant in the future though

nvie: OK, done in af3ee31d4198c03b3bc288570c68071d7ef7dbd2 — lmk what you think 👍

src/rules/noFlowFixMeInStrictFiles.js
```diff
@@ -0,0 +1,55 @@
...
+          });
+        }
+      }
+    },
+  };
+};
+
+export default {
+  create,
+};
```
Brianzchen, line 55: are you missing a schema?

